PR #521, opened 2012-07-19
Preserve return value in MiniTest example

mjackson:
The return value of MiniTest::Unit::TestCase is significant and needs
to be returned in order for the test reporter to work properly.

I also added an example for overriding the base implementation instead
of doing it in a subclass.


janko-m:
Doesn't `#transaction` always return the return value of the block? If yes, I think it's better to revert to the previous way, because it's less verbose (and it shows that Minitest's setup is just as pretty as RSpec's :smiley:).

jeremyevans:
I believe transaction currently returns the block status only if an exception is not raised, and :rollback=>:always raises a rollback exception internally.  Now that minitest-hooks is available, I should add an example that uses minitest-hooks and shows how you can get per-suite transactions with per-example savepoints.

Issue #718, opened 2013-10-10
uninitialized constant Sequel::Postgres (NameError)


jraines:
I have a bare bones sinatra app, with the following Gemfile:

gem 'sinatra'
gem 'pg'
gem 'sequel'
gem 'sequel_pg'

And this in config.ru

require 'pg'
require 'sequel'
require 'sequel_pg'
require './dashboard'

DB = Sequel.postgres 'my connection string'
run Sinatra::Application

And I get this error:

`uninitialized constant Sequel::Postgres (NameError)`

This is Ruby 1.9.3 and Sequel 4.3.0, pg 0.17.1

PR #720, opened 2013-10-10
Add note about explicit adapter requires

jraines:
I needed this to get past issue #718


elbouillon:
Note for later when using Bundle to autoload gems :

In config.ru

```
require 'rubygems'
require 'bundler/setup'
Bundler.require(:default)
```

In Gemfile

```
gem 'roda'
gem 'sequel'
gem 'sequel_pg', require: false
```

PR #819, opened 2014-05-21
Support Postgres "CREATE EXTENSION"

aschrab:


janko-m:
I don't quite understand the reasoning against this. It feels so natural to create extensions inside of migrations. It's lame that you would need to manually create extensions outside of migrations before you deploy. Why is it a bad practice to develop your application under a superuser account, if your database is only used by your application (which is most of the cases)? I mean, I don't see a problem, but I'm also not experienced in database security. I think if Heroku allows you to create extensions from migrations, then it's pretty much secure.

It would be really nice to have this. Everything in Ruby :), plus it's reversible.

jeremyevans:
Do you run your applications under the root operating system account?  Because using the database superuser account is basically the same idea.  Loading a C extension has security ramifications for the entire database cluster, not just the specific database you load the extension into, which is why it requires superuser access and not just ownership of the database.  Read the PostgreSQL documentation and see what their recommendations are.

Heroku does not allow you to create arbitrary extensions implemented in C, as that would be a huge security vulnerability. I believe they have a modified PostgreSQL installation that allows normal users to load extensions that have been specifically whitelisted.

If you want to do this all in ruby, just create an external Sequel extension that does this.  I'll even link to it from Sequel's plugin page. But I don't want to ship it with Sequel.

PR #933, opened 2015-01-15
Add support for PG citext columns under JRuby

rosenfeld:


jeremyevans:
This is specific to your PostgreSQL installation.  The oid for citext is not fixed (just as it isn't fixed for hstore or I assume most other extensions), so you can't hard code it.

Any change should be jruby/postegresql adapter specific, since I believe the postgres adapter currently handles this fine.

rosenfeld:
Ah, I see. Good to know, thanks. 

Anyway, I only sent the PR in case it was useful, as I already had the code as I was curious to test it. I've already finished my benchmarks on JRuby for now and it was a bit slower than MRI so I don't have any plans to move to JRuby. That means I don't intend to work on a correct patch for it currently since it doesn't affect my application particularly... But if I decide to move to JRuby it's good to know what would be the right direction to tackle this.

Thanks

PR #938, opened 2015-01-23
Store correct version with IntegerMigrator with allow_missing_migration_files

blerins:
Hi!

I have some issues with migrations with missing files. After migrating files 1 and 3, for example, the version number saved on database is 2, so if I try to upgrade the database migration 3 is run again.

This patch should resolve it.

Thank you!

Bruno.


jeremyevans:
Thanks for the patch!  This looks good.  I'll update the specs to test that the correct migration number is used.

PR #939, opened 2015-01-24
Document use of Sequel.extension :blank with :allow_blank

iseem:
The need for Sequel.extension :blank when using the :allow_blank validation helper is documented in the code comments, but not in the guide. So I added it.


jeremyevans:
Thanks for the patch!  This looks good.

PR #940, opened 2015-01-30
Add a retry_sleep option to Database#transaction

uhoh-itsmaciek:
This allows Sequel to pause between retries which can help with serialization conflicts on slow queries.

### DIFF
--- a/lib/sequel/database/transactions.rb
+++ b/lib/sequel/database/transactions.rb
@@ -45,6 +45,9 @@ module Sequel
     # :num_retries :: The number of times to retry if the :retry_on option is used.
     #                 The default is 5 times.  Can be set to nil to retry indefinitely,
     #                 but that is not recommended.
+    # :retry_sleep :: The number of seconds to sleep in between retries. Helpful on
+    #                 serialization errors for long-running queries, where retrying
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch.  This looks good.  Unfortunately, it's missed the cutoff for 4.19.0, so it will have to wait until 4.20.0.  I'll try merging this on Monday.

I think we may want to remove the `to_i` call, so you can set it to sleep for less than a second.  But I can take care of that after merging.

uhoh-itsmaciek:
No problem re: cutoff. And I agree on dropping the `to_i`.

jeremyevans:
I've given this some more thought, and think a more flexible idea would be a `:before_retry` option that took a proc that would be called with `num_retries` whenever a retry was performed.  You could use this for simple sleeping, for sleeping with exponential backoff, for extra logging, or for anything else.  Thoughts?

uhoh-itsmaciek:
Sure, that makes sense. I added the error that was raised as well--I think as long as we're going general, that can be useful. It's in #941.

jeremyevans:
Closing in favor of #941

PR #941, opened 2015-02-01
Add before_retry transaction option

uhoh-itsmaciek:
Like #940, but more general.


jeremyevans:
This looks good.  I'll try testing and merging on Monday.  Thanks for the patch!

jeremyevans:
I'm sorry that I didn't get to this today.  I should be able to take care of it tomorrow.  Apologies for the delay.

PR #945, opened 2015-02-14
Mysql2 does not use :charset, it use only :encoding.

tmtm:
When using mysql2 adapter, Sequel.connect :charset option does not work correctly.
Mysql2 does not use :charset, it use only :encoding.


jeremyevans:
Looks good, thanks for the patch!  I'll try merging soon.

PR #946, opened 2015-02-15
Add timezone support to swift adapter when using mysql

asppsa:
Hi there,

I've been having some trouble with the mysql2 gem so decided I would give the swift adapter a go as a possible alternative.  The main thing I noticed that didn't seem to be working was timezones for `DateTime` objects (i.e. for `DATETIME` columns in the db).  Regardless of what `database_timezone` and `application_timezone` settings I use, the `DateTime` objects always seem to get the UTC timezone.  I assume that this is only an issue for MySQL, but I haven't tested any other databases that Swift works with.    [this file](https://github.com/shanna/swift/blob/master/test/test_timestamps.rb) from their test cases seems to indicate that they only expect timezones to work with Postgres.

I've had a go at resolving this in the present pull request, but I don't think my solution isn't very good -- perhaps there is a better method that I could override?  Also, I had to use `#send` because `Sequel#convert_input_datetime_no_offset` is a private method.  This seems a bit hacky.

Finally, I haven't written any tests for this change because I wasn't sure where they should go.  It seems that there are no Swift tests at all at the moment -- is that right?

While I have the opportunity I'd also like to add that this is an awesome project!  Thanks for all your work on it.

### DIFF
--- a/lib/sequel/adapters/swift/mysql.rb
+++ b/lib/sequel/adapters/swift/mysql.rb
@@ -34,7 +34,21 @@ module Sequel
       class Dataset < Swift::Dataset
         include Sequel::MySQL::DatasetMethods
         APOS = Dataset::APOS
-        
+
+        def fetch_rows(sql)
+          super(sql) do |h|
+            converted_h = h.map do |k,v|
+              converted_v = if v.is_a?(DateTime)
+                              Sequel.database_to_application_timestamp(Sequel.send(:convert_input_datetime_no_offset, v, Sequel.database_timezone))
+                            else
+                              v
+                            end
+              [k, converted_v]
+            end
+            yield Hash[converted_h]
+          end
+        end
+
         private
         
         # Use Swift's escape method for quoting.


jeremyevans:
First, thanks for the patch.

This patch would cause a significant speed hit when using the swift/mysql adapter.  For better performance, you can directly modify the hash that is yielded, instead of creating a new hash.  Also, you should use `db.timezone` instead of `Sequel.database_timezone`.

You'll want to run the adapter tests with this patch to see if it fixes things.  You'll want to run the `spec_mysql` task, with the `SEQUEL_INTEGRATION_URL`  environment variable set to your swift/mysql connection string (e.g. `swift:mysql://...`).  The specs that fail on swift will be skipped by default, so you should run it first without this patch with the `SEQUEL_NO_PENDING` environment variable set and see which specs fail, then run it again with this patch with the `SEQUEL_NO_PENDING` environment variable set.  That should allow you to see which specs this patch fixes.

Assuming that this patch works as expected, there should be three specs in the `timezone_test.rb` file that this fixes, and possible other specs fixed.  The specs that this fixes should be unguarded, with `:swift` arguments removed from the related `cspecify` methods.

Note that the Sequel adapter is not the best place to try to fix this, which is why your fix is kind of hacky.  If the driver does typecasting (true for the Sequel swift adapters), then it is better to fix it in the driver.  If you could talk to the swift-db-mysql developers and see if they would be open to supporting `:database_timezone` and `:application_timezone` options, that would be best.

asppsa:
Great job - you totally beat me to it!  I will float the idea with the swift people of adding timezone support to their typecasting code.

PR #950, opened 2015-02-18
Merge arbitrary options into add_constraint ops hash

mpalmer:
Based on comments from Jeremy on sequel-talk, and the behaviour of most
other methods, my intuition is that add_constraint should merge any opts
passed to add_constraint.  However, the existing code wasn't doing that in
the case of a named constraint.  Now it does.


jeremyevans:
Unfortunately, your intuition has misled you.  Remaining arguments to `add_constraint` are used as filter arguments, and hashes are probably the most common type of filter arguments.  Consider:

```ruby
add_constraint :foo_id_c, :foo_id=>[1,2,3]
```

For `add_constraint`, options are passed via the first argument, which the current code already handles.

PR #951, opened 2015-02-20
Allow create_enum to be used in a reversible migration

celsworth:
As discussed on IRC, this PR adds the ability for create_enum in the pg_enum extension to be used in a reversible transaction.

I haven't added the spec yet, if you can let me know how to load the migration extension safely in pg_enum_spec.rb I will happily do that as well (or feel free to finish the job) :)


jeremyevans:
This implementation looks good, thanks for the patch!

I think you can just do `Sequel.extension :migration` in the pg_enum spec, then test the reversing similar to how the migration spec tests it.  After the spec gets added, I'll try merging and testing.

PR #955, opened 2015-02-24
Update documentation

agugaillard:
At least with sqlite3, the initial script from the README, can't be executed more than once.
This is occasioned when it opens a database that already contains a table with that name.
Even though this is documented, it is more appropriate to read it here too, the first contact with Sequel.


jeremyevans:
The example shows a memory database, which will always start out empty, so the existing code is correct.  I think the first code example in the README should be kept as simple as possible, and it's not the place to go over the different methods for creating tables.

PR #964, opened 2015-03-01
Add the #validates_confirmation helper

janko-m:
I am now experimenting with doing validations in service objects, so I really like that Sequel provides instance-level validation helpers. I was just missing `#validates_confirmation`.

I'm aware that users have to declare the confirmation column manually, but I think it's not a problem. What do you think?

### DIFF
--- a/doc/validations.rdoc
+++ b/doc/validations.rdoc
@@ -230,6 +230,18 @@ These methods check that the specified attributes can be valid integers or valid
     end
   end
 
+=== +validates_confirmation+
+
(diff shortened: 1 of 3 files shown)


jeremyevans:
`validates_confirmation` is specifically not present in validation_helpers, because it should not be a model level validation.  You wouldn't have two model columns that are the same thing, it's just something used to ensure that two form inputs are identical.  The validation belongs at the controller level, and not at the model level.

Note that the validation_class_methods plugin has `validates_confirmation_of` and `validates_acceptance_of` methods, so if you really want to validate confirmation (or acceptance) at the model level, you can use that plugin.

PR #966, opened 2015-03-09
Fix typo in docs

lucapette:
Spotted a small typo with reading the doc. Thank you for this awesome library!


jeremyevans:

PR #968, opened 2015-03-11
Demodulize default root name in json_serializer

janko-m:
I'm really loving the extension/plugin architecture of Sequel, thank you for making it that way.

I'm making a Sinatra app that is structured as a gem. So all classes, including Sequel models, are namespaced inside the main namespace. Currently Sequel does the following:

```rb
module Namespace
  class Album < Sequel::Model
  end
end

Namespace::Album.new.to_json(root: true) #=> "{"namespace/album": {}}"
```

I think a natural default key would be simply `"album"`. This goes along with the rest of Sequel where the namespace doesn't affect the name (e.g. table name). I know I could write `root: "album"`, but it gets tedious when you always need to pass it both to instance's and dataset's `to_json`.

I think this shouldn't be a backwards-incompatible change, because:

1. I believe it's a very rare case that someone even has a namespaced Sequel model
2. Users who wanted to get around this problem most probably are already passing `root: "album"`, or adding the root manually outside of the json_serializer extension.

### DIFF
--- a/lib/sequel/plugins/json_serializer.rb
+++ b/lib/sequel/plugins/json_serializer.rb
@@ -290,7 +290,9 @@ module Sequel
           end
 
           if root = opts[:root]
-            root = model.send(:underscore, model.to_s) unless root.is_a?(String)
+            unless root.is_a?(String)
+              root = model.send(:underscore, model.send(:demodulize, model.to_s))
+            end
             h = {root => h}
           end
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch.  This is a backwards incompatible change, but I think the risk is small enough that it's OK for a minor release, because it is unlikely someone is relying on the current behavior.

I'll try to get this tested and merged tomorrow.

PR #969, opened 2015-03-12
Improve error message of failed database extension

janko-m:
When I first got the error, I couldn't quite figure out how to fix the problem. I think this error mostly happens when an extension is required on the wrong level. This error happened to me because I did this:

```rb
DB.extension :pg_json, :pg_json_ops
```


jeremyevans:
Thanks for the patch.  I agree that this change makes the error more obvious.  I'll try to get this tested and merged tomorrow.

PR #971, opened 2015-03-12
edit rdoc

cookiebody:
Add demo code to make it clear how to use `pg_hstore_ops`.


jeremyevans:
Thanks for the patch!

PR #972, opened 2015-03-14
JDBC adapter should set username even if no password provided.

ecin:
**Scenario** 
Using JDBC to connect to MySQL under JRuby. The connection uses SSL certificates to authenticate users, thus it requires a username to be passed, but not a password.

**Issue**
The JDBC adapter only sets a username on the connection if a password is also set. 

There's no `jdbc_spec.rb` file, so not sure on the best way to test this change. Thoughts?

### DIFF
--- a/lib/sequel/adapters/jdbc.rb
+++ b/lib/sequel/adapters/jdbc.rb
@@ -201,7 +201,12 @@ module Sequel
           get_connection_from_jndi
         else
           args = [uri(opts)]
-          args.concat([opts[:user], opts[:password]]) if opts[:user] && opts[:password]
+          if opts[:user]
+            args.push(opts[:user])
+            if opts[:password]
+              args.push(opts[:password])
+            end
+          end
           begin
             JavaSQL::DriverManager.setLoginTimeout(opts[:login_timeout]) if opts[:login_timeout]
             JavaSQL::DriverManager.getConnection(*args)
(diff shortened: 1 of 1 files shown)


ecin:
Also of note: passing in an empty `:password` would have solved the problem for us. Would still prefer the `:username` and `:password` to not be coupled.

jeremyevans:
Thanks for the patch!

To test the jdbc/mysql adapter, you would run `spec_mysql` with the connection string environment variable set to `jdbc:mysql://...`.  Similar specs for other jdbc subadapters.  For jdbc subadapters that don't have a related database spec, `spec_integration` will work.

This patch looks fine to me.  I'll try to test it on Monday, and as long as it doesn't break anything (and it shouldn't), I can merge it then.

jeremyevans:
This patch doesn't actually work.  The JDBC API does not support providing a username but no password, see http://docs.oracle.com/javase/7/docs/api/java/sql/DriverManager.html.  So if you provide a username without a password with this patch, a `NameError` is raised by `getConnection`.  `NameError` is not rescued by the surrounding rescue statement, so the code you wrote for the properties is not reached.

Since you already have a workaround for this issue, I think it's better to just close this now. If you want to change this behavior, please have the JDBC developers add an API that allows for connecting with a username and not a password first.

PR #975, opened 2015-03-16
Add missing `super` statement to example.

frodsan:


jeremyevans:
Thanks for the patch.  Always a good idea to include super when overriding one of the methods defined by Sequel.

PR #978, opened 2015-03-17
add #becomes to sti models

isaiah:
This only implements `#becomes`, `#becomes!` is more complicated and will leave for future.

### DIFF
--- a/lib/sequel/plugins/single_table_inheritance.rb
+++ b/lib/sequel/plugins/single_table_inheritance.rb
@@ -215,6 +215,15 @@ module Sequel
       end
 
       module InstanceMethods
+        # convert to another type
+        def becomes(klass)
+          klass.call(@values)
+        end
+
+        def reload
+          _no_subclass_ds { super }
+        end
+
         private
 
         # Set the sti_key column based on the sti_key_map.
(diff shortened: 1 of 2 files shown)


jeremyevans:
Is there a reason we would want to add this?  It seems complicated and has questionable semantics.  I know ActiveRecord uses the API, but that doesn't make it a good idea.  Since you are just adding methods, you could always just add it as an external plugin that depends on the single_table_inheritance plugin.

isaiah:
Sometimes it is important that sti subclasses can change types, it could be done via monkey patch or plugins, but I think someone else might find it useful to be in the upstream.

jeremyevans:
Changing the STI class is currently simple:

```ruby
m = StiClass1.first
m.update(:kind=>'StiClass2')
StiClass2[m.id]
```

Is there really a need for all of the extra complexity?

isaiah:
This is different than `StiClass1.first.becomes(StiClass2)`, which only change the wrap class, thus the logic, but not the data, and there is no database operation. Talking about *complexity*, I think there are plenty of better candidates, e.g.

> Sequel doesn't directly support creating a bunch of associated objects and delaying saving them to the database until the main object is saved.

It leads to endless of hooks, but that's another story. Please close at your will.

Issue #980, opened 2015-03-24
Virtual rows' "Operators" section is a bit difficult to read


janko-m:
In the documentation for virtual rows, I find the ["Operators"](http://sequel.jeremyevans.net/rdoc/files/doc/virtual_rows_rdoc.html#label-Operators) section a bit difficult to understand. This is because the operators are written in the prefix form.

Since this documentation is introducing virtual rows, I think it would be better to switch the arguments (`1 - a` => `a - 1`), and use the infix form instead (and now you can maybe use the `=~` operator instead of hashes). At the end you could note that for `1 - a` they can use the prefix form, or they could also use `Sequel.expr(1) - a`, right?

I would normally write a pull request, but I'm usually pedantic about my own READMEs, so I assume you would prefer to write it yourself.

jeremyevans:
You need to understand why it is this way.  This section describes the virtual rows operator methods, which is why you have to use the prefix form with multiple arguments.

What you are describing is Sequel's general expression operator methods, which is documented elsewhere (like the dataset filtering guide).  What you are describing may be a better way in general, but it doesn't describe the virtual row-specific support, which is what this section attempts to do.

If you feel strongly that we should update the documentation, please send a pull request that adds additional examples to this section, without changing what is already there, showing the expression operator method alternatives to the virtual row operator methods.

PR #981, opened 2015-03-24
Note infix operators in virtual rows' rdocs

janko-m:
Related to #980.


jeremyevans:

PR #982, opened 2015-03-24
Add refile-sequel to list of external extensions

janko-m:
I recently made an [extension](https://github.com/refile/refile-sequel) for [Refile](https://github.com/refile/refile) to integrate it with Sequel.


jeremyevans:

PR #986, opened 2015-04-07
Fix 'undefined method error' in #to_hash_groups

stevenringo:
I came across a NoMethodError: undefined method `values_at' for #<Redacted:0x007fc89fcc0550>, when trying to use this method.

Hoping this fixes it, as my use case only covered the code on line 741.


jeremyevans:
This would break virtually all uses of the method (did you try running the tests?).  Line 732 calls naked on the dataset, which should remove the row_proc from the dataset and cause it to yield plain hashes.  

Please post a self contained example showing the problem you are having to the sequel-talk Google Group, and we can try to help over there.

PR #987, opened 2015-04-09
wishdev fork is stalled , rowland is maintained

mariuz:
replaced with https://github.com/rowland/fb

I will check if there we can merge the old wishdev branch

### DIFF
--- a/lib/sequel/adapters/firebird.rb
+++ b/lib/sequel/adapters/firebird.rb
@@ -3,7 +3,7 @@ Sequel.require 'adapters/shared/firebird'
 
 module Sequel
   # The Sequel Firebird adapter requires the ruby fb driver located at
-  # http://github.com/wishdev/fb.
+  # https://github.com/rowland/fb.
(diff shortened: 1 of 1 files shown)


jeremyevans:
Rowland's repository isn't a fork, that's the main repository.  If you look at https://github.com/wishdev/fb, it says right at the top that it was forked from rowland/fb.

Sequel's firebird adapter doesn't work with the standard fb driver last I checked. wishdev's fork has 15 commits to change fb to make it work with Sequel's firebird adapter, but as far as I'm aware, it changes the fb API and would not be appropriate for merging upstream.

I'm closing this now.  If you can provide test output (`rake spec_firebird`) using both wishdev's fork and the standard fb driver that shows the standard fb driver passes more tests, then I'll be happy to reopen and merge this.  However, this is probably a moot point, because unless most of the tests start passing, I plan to deprecate the firebird adapter in the next release and remove it in a future release.

Issue #577, opened 2012-10-30
Export dataset to CSV using ruby's standard library.


marcalc:
This helps to fix a lot of potential problems like encoding, and commas inside strings.

``` ruby
class Sequel::Dataset
  require 'csv'
  def to_csv(include_column_titles = true)
    n = naked
    cols = n.columns
    csv_string = CSV.generate do |csv|
      csv << cols if include_column_titles
      n.each{|r| csv << cols.collect{|c| r[c] } }
    end
    csv_string
  end
end
```

Thanks for this great gem.

PR #988, opened 2015-04-09
csv_serializer plugin

bjmllr:
A plugin for working with CSV in the spirit of `json_serializer` and `xml_serializer`. The API is broadly similar to the existing serialization plugins, but because CSV is not a nested format, no attempt was made to support associations. It will of course properly escape values, accept alternative record separators, and so on, as it uses the standard `csv` library (or, on ruby < 1.9, `fastercsv`).

Past interest in CSV support:
* https://github.com/jeremyevans/sequel/issues/577
* https://groups.google.com/forum/#!searchin/sequel-talk/csv/sequel-talk/qq5OiXkn8LM/GKnFtR2VS84J
* https://groups.google.com/forum/#!searchin/sequel-talk/csv/sequel-talk/pUcaBSRc1_A/yLg21y49DoMJ
* http://stackoverflow.com/questions/24763711/reading-a-csv-file-into-ruby-using-sequel-and-the-sqlite3-gems

`fastercsv` greatly simplified the implementation for ruby 1.8, but isn't used by later rubies (in fact it throws an exception telling you to use the standard library if you try to use it), so I added it only to `.travis.gemfile` and not `sequel.gemspec`, and arranged for it to be conditionally loaded when the plugin is loaded.

Please let me know if there are any ways this plugin can be improved.

Thanks!

### DIFF
--- a/.travis.gemfile
+++ b/.travis.gemfile
@@ -15,6 +15,10 @@ if RUBY_VERSION < '1.9.3'
   gem "i18n", '<0.7.0'
 end
 
+if RUBY_VERSION < '1.9.0'
+  gem "fastercsv"
+end
+
 # MRI/Rubinius Adapter Dependencies
 platforms :ruby do
   gem "sqlite3"
--- a/lib/sequel/plugins/csv_serializer.rb
+++ b/lib/sequel/plugins/csv_serializer.rb
@@ -1,0 +1,212 @@
+if RUBY_VERSION < "1.9"
+  begin
+    require 'fastercsv'
+  rescue LoadError
+    warn("plugin csv_serializer requires either ruby>=1.9 or fastercsv")
+  end
+else
+  require 'csv'
+end
+
+module Sequel
+  module Plugins
+    # csv_serializer handles serializing entire Sequel::Model objects to CSV,
+    # as well as support for deserializing CSV directly into Sequel::Model
+    # objects.  It requires either the csv standard library from ruby >= 1.9,
+    # or the fastercsv gem.
+    #
+    # Basic Example:
+    #
+    #   album = Album[1]
+    #   album.to_csv(:write_headers=>true)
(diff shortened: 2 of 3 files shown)


jeremyevans:
From a first glance, this looks really good, thanks for the patch.  I'm definitely in favor of including this with Sequel.  I probably won't have time to do a more thorough review until Monday, but I definitely expect to get this merged sometime next week.

bjmllr:
Thanks for accepting this and cleaning it up. I've been studying 8ad8718e40851613e5089b63f6cc6091d7898188 and I have a couple questions:

In the spec, I see that the example models `Artist` and `Album` were changed to anonymous classes. Is this a change that could be applied to the XML and JSON serializer specs as well? Do those specs need to continue using named classes?

In light of the recent change to `CONTRIBUTING`, and considering the scope of the cleanup here, would it have been better to request feedback on the patch via sequel-talk before opening the PR? I felt that the code as submitted was ready to be used (I'm using that version now while I wait for the release of 4.22), but I did sense that there would be some cleanup, should that be considered ready to merge?

jeremyevans:
In general, it's best not to modify global state unless it is necessary, and adding a constant modifies global state. So for most newer specs, I prefer the anonymous class approach.  I don't think it's worth changing the existing specs for the other plugins, though.

In this case, the pull request you submitted was very good, well tested, and already closed mirrored the existing xml and json serializer plugins, so I didn't think any discussion was necessary.  Most of the changes I made were stylistic or designed to simplify the plugin, as opposed to changes in how the plugin should behave.  Generally, discussions on sequel-talk are necessary to determine desired behavior, but that wasn't really needed in this case.

It's never a bad idea to ask for feedback on sequel-talk, especially if you have questions about how something should work or the best way to integrate it.  But in this particular case, I didn't think it was necessary.

Thanks again for the patch!

bjmllr:
Thanks for the explanation, makes perfect sense.

Issue #989, opened 2015-04-10
Allow ruby defaults on Postgres array columns


schwern:
It would be nice if one could specify defaults on Postgres array columns as Ruby arrays.

    column :titles, 'varchar[]', default: []

This solves two problems.  First, `default: []` is currently ignored with no warning.  The column will wind up with no default.

Second, if it has a not null constraint auto_validations will not see it has a default and do a presence validation.  auto_validations looks only for a ruby_default (possibly a bug).  This makes it difficult to use auto_validations with Postgres arrays.

I'd implement it, but I'm not sure where it would go.

jeremyevans:
This isn't a bug.  When you do:

```ruby
column :titles, 'varchar[]', default: []
```

Sequel runs the SQL:

```sql
CREATE TABLE "a" ("titles" varchar[] DEFAULT (NULL))
```

So it is not ignoring the `:default` option.  Ruby arrays are literalized as SQL value lists by default, and the empty array is literalized as the value list `(NULL)`, which results in a `NULL` default for the column.

If you want to use an empty PostgreSQL array as the default, do `:default => Sequel.pg_array([])`.  This requires you load the pg_array extension into the Database object. I think if you fix your `:default` option value, your problem will be solved.

schwern:
I do not get that behavior.  This code...

    column :address, 'varchar[]', default: []

Produces this SQL

    address character varying[]

Furthermore, I get an error if I try to use an array with elements in it.

      column :address, 'varchar[]',     default: ["foo", "bar"]

Produces...

    "address" varchar[] DEFAULT ('foo', 'bar')

Which errors.

    Sequel::DatabaseError: PG::DatatypeMismatch: ERROR:  column "address" is of type character varying[] but default expression is of type record

I am using these extensions.

    Sequel::Model.db.extension :pg_array, :pagination, :pg_json, :pg_location

I do not agree that it should be null.  `[]` and `nil` are not the same in Ruby.  `{}` and `NULL` are not the same in a PG array column.  If I meant to pass NULL I will use nil or just no default.

From a functional perspective, this leaves no way to have an empty array as a default for a PG array column.

PG array columns should either turn Ruby arrays into PG arrays, or they should error when they see an array.  The current behavior seems like its just the default behavior for scalar columns.

jeremyevans:
This code...

```ruby
column :address, 'varchar[]', default: []
```

Produces this SQL:

```sql
[...]

PR #990, opened 2015-04-10
Don't call validate_presence on not null columns with non-ruby defaults.

schwern:
For example, this will have a default, but not a ruby_default (though it probably should, see #989).  auto_validates was only looking for ruby_default.

    column :titles, 'varchar[]', null: false, default: '{}'

### DIFF
--- a/lib/sequel/plugins/auto_validations.rb
+++ b/lib/sequel/plugins/auto_validations.rb
@@ -109,7 +109,10 @@ module Sequel
 
         # Parse the database schema and indexes and record the columns to automatically validate.
         def setup_auto_validations
-          not_null_cols, explicit_not_null_cols = db_schema.select{|col, sch| sch[:allow_null] == false}.partition{|col, sch| sch[:ruby_default].nil?}.map{|cs| cs.map{|col, sch| col}}
+          not_null_cols, explicit_not_null_cols =
+                 db_schema.select{|col, sch| sch[:allow_null] == false}
+                   .partition{|col, sch| sch[:default].nil? && sch[:ruby_default].nil?}
+                   .map{|cs| cs.map{|col, sch| col}}
           @auto_validate_not_null_columns = not_null_cols - Array(primary_key)
           explicit_not_null_cols += Array(primary_key)
           @auto_validate_explicit_not_null_columns = explicit_not_null_cols.uniq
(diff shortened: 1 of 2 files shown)


jeremyevans:
The problem with this is that `:default` on some adapters will be non-nil even if the default value for the database is NULL.  Basically, the `:default` schema value is adapter dependent (like `:db_type`), it shouldn't be relied on as you are doing here.

Unfortunately, some columns with database defaults are not representable by ruby defaults.  For those columns, you have to add exceptions to the default auto_validations behavior, see the auto_validations plugin for details.

In your particular case of a PostgreSQL array, it's possible that support could be added in the pg_array extension to recognize default values for array columns, so that `:ruby_default` could be populated.  So if you want to try to fix the underlying issue, you should override `Database#column_schema_to_ruby_default` in the pg_array extension to handle array columns appropriately.

schwern:
> The problem with this is that :default on some adapters will be non-nil even if the default value for the database is NULL. Basically, the :default schema value is adapter dependent (like :db_type), it shouldn't be relied on as you are doing here.

Shouldn't this be considered a bug in the adapters?

> You should override Database#column_schema_to_ruby_default in the pg_array extension to handle array columns appropriately.

Thank you, that gives me something to hook into.

jeremyevans:
No, it is not considered a bug in the adapters.  Just like the difference between `:db_type` and `:type`,  `:default` is what the underlying driver provides as the default value, which can mean different things depending on the driver used.  `:ruby_default` is similar to `:type`, a normalized version that should be adapter independent and safe to use.

schwern:
[The Sequel::Database::schema() docs](http://sequel.jeremyevans.net/rdoc/classes/Sequel/Database.html) do not communicate any of that subtlety by my read.  They are "database specific", but nothing that suggests they may be fabricated by the adapter and not actually exist in the database schema.  I'd like to patch the docs to convey that, but I'm struggling with how.  Would it be enough to say that `:db_type` and `:default` may not truly reflect the state of the database schema?

jeremyevans:
`:db_type` and `:default` are not "fabricated by the adapter", they are designed to be the direct output of the underlying driver, unmodified by the adapter.  For example, one driver may return `nil` for NULL defaults, another may return `''`, and a third may return `'NULL'`.

I think the only way to get what you want would be to change the Database adapter API so that `:default` is set to nil if the adapter can determine that NULL is the default value in the Database.  That's a possible change to make, but I think it would require adding an integration test to check for the behavior, and then checking all adapters that ship with Sequel to make sure they handle that new requirement correctly.  We'd also need to make special mention of this in the release notes so that external adapter authors are appropriately notified.

With that change made, then auto_validations could just check for a `:default` value, and could ignore the `:ruby_default`, fixing cases where a default is exists but the adapter does not parse it into a `:ruby_default`.

What are your thoughts on this?

PR #991, opened 2015-04-11
Give PG array columns ruby_defaults

schwern:
*This is not ready to be merged*

As discussed in #990, this is a first cut at giving PG array columns with defaults a ruby_default in the schema.  This fixes auto_validations and maybe other problems.

I would like some advice on how to proceed.

```
  1) pg_array extension should set a ruby_default for array types
     Failure/Error: ).should == @m::PGArray.new([1,2,3], :integer)
       expected: [1, 2, 3]
            got: ["1", "2", "3"] (using ==)
     # ./spec/extensions/pg_array_spec.rb:398:in `block (2 levels) in <top (required)>'
```

It seems I need the JSONCreator for integers.  Which creator to use only appears to be available in the [big list of register() calls](https://github.com/jeremyevans/sequel/blob/master/lib/sequel/extensions/pg_array.rb#L527) and I can't figure out a way to get at that information.

What should I be doing instead of `Creator.new(type).call(default)`?  Do I need to make PGArray::DatabaseMethods::register store the creator for each array_type in a structure for later use?  Or am I overlooking something?

### DIFF
--- a/lib/sequel/extensions/pg_array.rb
+++ b/lib/sequel/extensions/pg_array.rb
@@ -168,6 +168,7 @@ module Sequel
         typecast_method_map[db_type] = :"#{type}_array"
 
         define_array_typecast_method(mod, type, creator, opts.fetch(:scalar_typecast, type))
+        define_array_create_method(mod, type, creator)
 
         if oid = opts[:oid]
           type_procs[oid] = creator
@@ -176,6 +177,15 @@ module Sequel
         nil
       end
 
+      def self.define_array_create_method(mod, type, creator)
+        mod.class_eval do
+          meth = :"create_#{type}_array_from_default"
+          define_method(meth){|v| creator.call(v) }
+          private meth
+        end
+      end
+      private_class_method :define_array_create_method
+      
       # Define a private array typecasting method in the given module for the given type that uses
       # the creator argument to do the type conversion.
       def self.define_array_typecast_method(mod, type, creator, scalar_typecast)
@@ -307,7 +317,8 @@ module Sequel
         def column_schema_default_to_ruby_value(default, type)
           case type
           when *ARRAY_TYPES.values
-            Creator.new(type).call(default)
+            meth = "create_#{type}_from_default"
+            send(meth, default)
           else
             super
           end
--- a/lib/sequel/extensions/pg_array.rb
+++ b/lib/sequel/extensions/pg_array.rb
@@ -300,6 +300,19 @@ module Sequel
           end
         end
 
+        def column_schema_default_string_type?(type)
+          ARRAY_TYPES.values.include?(type) || super
+        end
+
+        def column_schema_default_to_ruby_value(default, type)
+          case type
+          when *ARRAY_TYPES.values
+            Creator.new(type).call(default)
+          else
+            super
+          end
+        end
+
         # Given a value to typecast and the type of PGArray subclass:
         # * If given a PGArray with a matching array_type, use it directly.
         # * If given a PGArray with a different array_type, return a PGArray
--- a/spec/extensions/pg_array_spec.rb
+++ b/spec/extensions/pg_array_spec.rb
@@ -397,4 +397,14 @@ describe "pg_array extension" do
              "'{1,2,3}'::integer", :integer_array
             ).should == @m::PGArray.new([1,2,3], :integer)
   end
+
+  it "should reject garbage array defaults" do
+    @db.send(:column_schema_to_ruby_default,
+             "'{1,2,3'::integer", :integer_array
+            ).should == nil
+
+    @db.send(:column_schema_to_ruby_default,
+             "'{foo,bar'::character varying", :varchar_array
+            ).should == nil
+  end
 end
--- a/spec/extensions/pg_array_spec.rb
+++ b/spec/extensions/pg_array_spec.rb
@@ -387,4 +387,14 @@ describe "pg_array extension" do
     @db.schema_type_class(:banana_array).should == Sequel::Postgres::PGArray
     @db.schema_type_class(:integer).should == Integer
   end
+
+  it "should set a ruby_default for array types" do
+    @db.send(:column_schema_to_ruby_default,
+             "'{foo,bar}'::character varying", :varchar_array
+            ).should == @m::PGArray.new(['foo', 'bar'], :varchar)
+
+    @db.send(:column_schema_to_ruby_default,
+             "'{1,2,3}'::integer", :integer_array
+            ).should == @m::PGArray.new([1,2,3], :integer)
+  end
 end


schwern:
I went ahead and implemented the self.register hack to make the creator information available as a pile of methods.  I don't like it, but it works.

jeremyevans:
On PostgreSQL, the schema information includes the oid for the column.  This information is not passed to `column_schema_to_ruby_default` though.  So it may be better to override `Database#schema_parse_table` in the `pg_array` extension, and have it iterate over the values returned by super to populate `:ruby_default` for array columns.  This will require changing the general`Database#schema` to not override `:ruby_default` if it is already present.  That should hopefully avoid the need to add the methods you've added in this pull request.

schwern:
It occurs to me that this suggests a Creator factory is needed, or a Typemap class.

jeremyevans:
I looked into this and it isn't possible currently and would require a completely separate parser for the defaults.  From your code, it appears you are under the impression that PostgreSQL returns default values the same way it would return general array values.  Unfortunately, that is not the case. PostgreSQL returns array defaults values in the following format:

    ARRAY[]::integer[] # []
    ARRAY[1] # [1]
    ARRAY[ARRAY[0, 1, 3]] # [[0, 1, 3]]

Sequel doesn't currently have a parser for this format, since it is not the format that PostgreSQL uses for outputting array values.  I believe the way PostgreSQL outputs defaults, Sequel would basically need to implement an SQL parser to parse it, since PostgreSQL is outputting an SQL expression.  Part of Sequel's philosophy is that it does not parse SQL.  So I don't see this as workable in the general case.

If you really think this should be fixed, you should work on a separate extension that can parse this format.  I think this should be in a separate extension, as few pg_array users would need it, and I don't think the rest should have to pay the memory penalty for a parser they won't be using.

schwern:
For not being possible, it sure works well.  Did you try the patch?

schwern:
Here is a little program demonstrating the PR works.

```
require 'sequel'

DB = Sequel.postgres('test', host: 'localhost')
DB.extension :pg_array

DB.create_table!(:check_arrays) do
  primary_key :id
  column :strings, 'varchar[]', null: false, default: '{foo, bar}'
  column :numbers, 'numeric[]', null: false, default: '{1.1, 2.2}'
end

class CheckArray < Sequel::Model
end

puts CheckArray.db_schema[:strings].inspect
puts CheckArray.db_schema[:numbers].inspect
```

jeremyevans:
PostgreSQL will use the same format for defaults as provided when the table was created.  Your examples only work because you are providing the literals as a string that is already in PostgreSQL array format.  Sequel literalizes arrays using the `ARRAY` format:

```ruby
DB.extension :pg_array
DB.create_table(:a){column :a, 'integer[]', :default=>[]}
DB.schema(:a)[:default]
# => "ARRAY[]::integer[]"
```

Since the parser cannot handle the `ARRAY` format, I think I was right that this approach cannot handle default parsing for arrays in general and the pull request should be closed.

With Sequel, it is best to not submit pull requests until they are ready to be merged.  If you want to discuss a potential new feature, please do so on the sequel-talk Google Group.  I know GitHub's approach is that pull requests are for starting conversions, but this is not how Sequel operates.  Discussion/support occurs on the Google Group, the GitHub issue tracker is for bugs only, and pull requests are only for code ready to be merged.

With the changes I made to auto_validations, it now uses the `:default` option instead of the `:ruby_default` option, so parsing of array column defaults is no longer necessary for correct behavior.  Not to mention that auto_validations supports manually overrides, so even in older versions of Sequel, you can adjust how the validations are done.

schwern:
This PR was [exactly what you suggested I try to do](https://github.com/jeremyevans/sequel/pull/990#issuecomment-91473276).  Then, after four hours of volunteer work, I'm told it's the wrong way to do it. :(  This is discouraging.

I don't want to join another mailing list just so I can get some discussion about my patches.  I certainly don't have the time to hang around on IRC.  Contributing is a when-I-have-time thing.  Project mailing lists and IRC have a lot of noise and chatter that I don't need to sift through.  Github PR's keep things nice and focused and compartmentalized.  It wasn't like I submitted garbage, it had tests and worked and I was willing to put in the effort, I just wanted some feedback on the implementation and what direction I should go.   Please consider allowing contribution review in PRs; they are designed for that purpose and very convenient for casual contributors.

I'm happy that #990 is fixed.  I don't mind that you did it in a way that [you said was no good and closed](https://github.com/jeremyevans/sequel/pull/990#issuecomment-91473276); everybody's allowed to be wrong.  But you did it without attribution or acknowledgement that I did it that way first, and that you said it was wrong and led me off on a wild goose chase.  Because the issue was closed I didn't even get a notification it was fixed.  This is all discouraging.

I feel like I'm getting a run around and my work is in vain.  On the whole I'm disinclined from spending more time contributing to this project, which is a shame because it's very good and I enjoy using it.  I'd rather push my fixes back to the main project than maintain my own branch.

Hopefully we can come up with some way to make this work.  I don't mind if you want to be Closey McCloserson and close issues and PR's early, but with the understanding that it's more about keeping your queue tidy.  I'm going to keep talking if I think there's more to be said and not take it like a door slamming in my face.  I also want to use PRs for "hey this works, but I need some eyeballs on it".  Again, you can close it if you like during that process, so long as that's more about keeping the queue tidy than a rejection.

If that's ok with you, we can work together.  If not, either suggest something else or I'll be on my way.

jeremyevans:
I'm sorry, I probably should have handled things better.

I suggested you do it this way because if Sequel could support it, this would be the way to do it.  Unfortunately, it turns out that because of how PostgreSQL formats defaults, this way would not work without writing a completely separate parser.  I did not know that when I originally made the suggestion, or I would not have suggested it.  I apologize that my ignorance wasted four hours of your time.

With what I know now, implementing this would be a lot of work for a fairly minor feature, so it doesn't seem worth it. If someone decides to do the work, it should be an a separate extension so the vast majority of users who don't need it don't have to load it.  That would be a very different implementation than the one in this pull request, which is why I closed this.

You don't have to join a mailing list to discuss Sequel, sequel-talk can be used via the web without ever getting email (that's how I use it).  It may require a little more work to use sequel-talk instead of GitHub to discuss Sequel for casual contributors.  However, it is more visible for Sequel community members, and that's the tradeoff I chose.  Sequel used sequel-talk before it used GitHub (before GitHub even exited I think).

Regarding #990, your pull request was originally rejected because at the time it was made it relied on unspecified adapter behavior.  I had to change the adapter API so that the behavior was specified, so that auto_validations could then start relying on it.  As I mentioned, I originally thought that actually parsing the default would be better, but that turned out not to be simple as I originally envisioned it would be.

I didn't use your implementation for #990, though naturally due to the change required my implementation ended up being similar to yours.  I did reference the pull request in the commit message and CHANGELOG, but if you'd like me to mention your username in the CHANGELOG as well, I'd be happy to do that.

I don't want to discourage you from contributing to Sequel.  However, I do ask in the future that you discuss changes to Sequel on the sequel-talk Google Group.  Only file issues in GitHub if they are actual bugs in Sequel, and only file pull requests that are ready to be merged.  I realize that may be different than how other projects use GitHub, but it has worked well for Sequel.

Please don't file pull requests for "hey this works, but I need some eyeballs on it".  The term pull request is taken literally, in the sense that you are requesting the code be pulled.  If the code isn't ready to be pulled, a request to pull is not appropriate.

If you'd like me to review your code but it isn't ready for merging, post on the sequel-talk Google Group, pointing to the branch or diff you would like reviewed, and any comments/questions you have.  That way I and other Sequel community members are informed about it.  I'll certainly take the time to review your code and give you feedback, if you have any doubt, feel free to look at the sequel-talk archives.

I may be too trigger happy with the close button, I admit that.  I generally close issues and pull requests as soon as I think that the issue is fixed or the pull request will not be merged.  That does not mean that further discussion on the issue/pull request is closed.  Often, there will be further discussion, and occasionally, this will result in the reopening of the issue/pull request.  Closing a pull request should not be taken as slamming the door, just that it appears unlikely the pull request will be merged.  As you guessed, one of the main reasons for this is keeping the queue tidy, to avoid the broken window effect.

I hope this helps you understand why the issues and pull requests were handled the way they were, and I hope that you will still want to contribute to Sequel in the future.

schwern:
> Your examples only work because you are providing the literals as a string that is already in PostgreSQL array format. 

Wait a minute.  Didn't I propose [allowing Ruby arrays for Postgres arrays](https://github.com/jeremyevans/sequel/issues/989) but [you rejected it](https://github.com/jeremyevans)?  Then I was supposed to support that rejected feature?

[Your example](https://github.com/jeremyevans/sequel/pull/991#issuecomment-92982971) only works the way it does because you changed the code out from under this PR in exactly the way you said Sequel wouldn't support.  I'm glad [you decided to implement it](https://github.com/jeremyevans/sequel/commit/a42092e02ed0a7070825596c6e080606b795cd18), and I do feel it is the better fix, but you didn't communicate it back to me.  The only reason I did this PR was because you didn't want to support ruby array defaults.  I feel even more that I've been run in circles.

I appreciate your response.  There's obviously a process and communication breakdown.  Right now I'm too annoyed to properly consider what to do next.

jeremyevans:
As I explained when I reopened that issue, I added support for using ruby arrays as PostgreSQL arrays in `:default` values despite misgivings that it would lead to people believing something that isn't true (that ruby arrays can be used as PostgreSQL arrays generally).  Basically, you requested something I thought was a bad idea, so I rejected it, you continued to ask for it, and despite the issues it causes, I put aside my judgement and decided to trust yours and implement it anyway.  I'm not sure why that makes you feel like you have been run in circles, but I'm sorry you feel that way.

I'm not sure why you weren't notified when I reopened the issue, you should have been since you opened the issue, unless you specifically changed your notification settings for the issue.  This is assuming I understand correctly how GitHub notifications work in this case, which I admit I may not. Was there some other way you wanted me to communicate the decision back to you?

Note that even if I didn't change Sequel to support ruby arrays as PostgreSQL arrays in `:default` values, that wouldn't change this particular issue, since any user could have used the `ARRAY` format when setting the default for the column. The recommended way before would have been to use `:default=>Sequel.pg_array([], :type)`, which would have resulted in the same issue.

Note that if we were using sequel-talk to discuss these issues, it is unlikely you would have any of these problems in the first place.  You could have opened up the initial discussion there, we could have discussed it there, and avoided all of these problems.  Plus it is more likely we would get feedback from the other members in the Sequel community.

Again, I'm sorry for how this situation has been handled.  I've updated the contribution guidelines to provide more information and hopefully avoid situations like this in the future.

schwern:
Ok.  I'll stop being a miserable old bastard and try the next patch your
way. :)

PR #999, opened 2015-04-28
Change single and class to use Plugins.inherited_instance_variables

QuinnHarris:
This is intended to clean up the single and class table inheritance plugins a little bit by using inherited instance variables.  This passes specs and I don't expect it will change behaviour.  This is an aesthetic change so I don't mind if its rejected.


jeremyevans:
This looks fine, I'll test and merge it soon.  Thanks for the patch!

PR #1001, opened 2015-05-03
List pagination methods in the documentation

janko-m:
Hi Jeremy,

I was thinking that it would be useful to list the methods that the pagination extension provides in the documentation. Just so that it's not necessary to open the source code.


jeremyevans:
Thanks for the patch.  This looks good, I'll try to merge later today.

PR #1002, opened 2015-05-03
this small patch makes add_named_conversion_proc also work for enum types

celsworth:
This makes adding a named conversion proc work for an enum type column.


jeremyevans:
Thanks for the patch, this looks good. We should probably add a spec for it, that adds an enum type, then calls `add_named_conversion_proc`, then checks that the conversion proc is called when a value of the enum type is returned by a query.

PR #1006, opened 2015-05-12
add reference to sequel-pg_advisory_locking

celsworth:


jeremyevans:
Thanks for the patch.  Looks good except for the omission of the closing li tag.

PR #1007, opened 2015-05-13
Add link to sequel-jsonapi_eager external plugin

janko-m:
I created a plugin that helps eager loading associations when exposing objects in a JSON API.


jeremyevans:
Thanks for the patch, will commit shortly.

PR #1010, opened 2015-05-15
Fix documentation of #full_text_search

janko-m:
Just and indentation fix, because it currently looks like this:

screen shot 2015-05-15 at 14 15 13


jeremyevans:
Good catch, thanks!

PR #1011, opened 2015-05-19
Start running tests against JRuby 9k on Travis

deepj:


jeremyevans:
Thanks for the patch.  This looks fine.  I've actually been testing with JRuby 9.0.0.0-pre1 locally since it came out (but not pre2 since that broke things).

headius:
@jeremyevans I just fixed jruby/jruby#2957 but I'm not sure what breakage you're seeing in pre2. Have you filed any issues I can look at?

jeremyevans:
@headius Sorry, I haven't.  Shortly after pre2 was released, there was talk about it breaking sequel in #jruby on IRC, so I thought you were aware of it.  Here's a link to the breakage Travis shows with jruby-head: https://travis-ci.org/jeremyevans/sequel/jobs/63142621.  I could post an issue on the JRuby bugtracker, but I haven't isolated the cause yet, so it wouldn't be more detailed than just pointing to the failure on Travis.

Also, I only tested pre1 and pre2 locally on Windows, not sure if that is related.

headius:
Ok, I will have a look at Travis and try to reproduce locally. Trying to
knock down all remaining user issues with 9k before we do RC1.

- Charlie (mobile)

deepj:
@headius I think the first issue on Travis having something to do with JRuby 9k-head because I wasn't be able to reproduce it on my machine with JRuby 9k-pre2. But I use Java 8 instead of Java 7 used on Travis. On the other hand I'm not able to install JRuby 9k-head via `ruby-build` though. This might be a bit useful for you.

PR #1013, opened 2015-05-23
Add missing sproc option

flash-gordon:
Hello,

I think there is missing option in StoredProcedureMethods#execute_insert that leads to fail for insert-like procedures when using JDBC adapter. I faced with it on Oracle but add a test for PG (also failed).

Best regards,
Nikita


jeremyevans:
Thanks for the patch!  This looks good to me.  Could you move the test to the postgres adapter tests (spec/adapters/postgres.rb), since it is already limited to only running on PostgreSQL?  I should be able to merge and test this on Tuesday.

flash-gordon:
Sure, done.

PR #1014, opened 2015-05-25
Reversible triggers and functions

celsworth:
This change makes `create_trigger` and `create_function` reversible.

Also a separate commit for adding some omitted methods in the list of change-safe methods.

### DIFF
--- a/doc/migration.rdoc
+++ b/doc/migration.rdoc
@@ -71,7 +71,9 @@ it will attempt to create a +down+ block for you, assuming that it knows how to
 reverse the given migration.  The +change+ block can usually correctly reverse
 the following methods:
 
+* +create_function+
 * +create_table+
+* +create_trigger+
 * +add_column+
 * +add_index+
 * +rename_column+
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch.  Unfortunately, I'm against this for a few reasons.

First, one instance where this actually does the wrong thing is when `:replace` is used as an option to `create_function`.  It's not possible to handle that case correctly in the reverser.  Since all triggers are based on functions, if you can't make `create_function` reversible, it's probably a bad idea to make `create_trigger` reversible.

Second, I think it's a bad practice to add adapter-specific method handling to the generic migration reverser.

Third, `create_trigger` and `create_function` really only exist for backwards compatibility, I think it was a mistake to add them, and I don't encourage their usage in new code.  Database specific code like that should just use raw SQL with `Database#run`.  The only reason I didn't remove these methods in Sequel 4 is that I didn't want to break old migrations.

I'll cherry-pick the second commit, since you are right that `create_join_table` and `create_view` should be listed.  Adding support for `create_view` was probably a bad idea, as on some databases that doesn't handle `:replace` correctly either, though the recommended usage for that would be to call `create_or_replace_view` instead of `create_view`.

Issue #1018, opened 2015-06-11
Getting "outside valid range for the datatype INTEGER” on insert when auto increment primary key is large BIGINT


DougEverly:
```ruby
sql = "INSERT INTO table (pk, name) VALUES (NULL, 'Doug')”
pk = DB[sql].insert
```

When the auto-incrementing pk is "148171983513" BigInt for example.

The row gets inserted correctly, but seems the primary key return value is expected to be Int, not possible BigInt.

```
Sequel::DatabaseError: Java::ComMysqlJdbcExceptionsJdbc4::MySQLDataException: '1.48171983512E11' in column '1' is outside valid range for the datatype INTEGER.
       raise_error at /home/deverly/my_project/vendor/cache/jruby/2.2.0/gems/sequel-4.23.0/lib/sequel/database/misc.rb:423
         statement at /home/deverly/my_project/vendor/cache/jruby/2.2.0/gems/sequel-4.23.0/lib/sequel/adapters/jdbc.rb:653
  block in execute at /home/deverly/my_project/vendor/cache/jruby/2.2.0/gems/sequel-4.23.0/lib/sequel/adapters/jdbc.rb:246
              hold at /home/deverly/my_project/vendor/cache/jruby/2.2.0/gems/sequel-4.23.0/lib/sequel/connection_pool/threaded.rb:103
       synchronize at /home/deverly/my_project/vendor/cache/jruby/2.2.0/gems/sequel-4.23.0/lib/sequel/database/connecting.rb:255
           execute at /home/deverly/my_project/vendor/cache/jruby/2.2.0/gems/sequel-4.23.0/lib/sequel/adapters/jdbc.rb:245
    execute_insert at /home/deverly/my_project/vendor/cache/jruby/2.2.0/gems/sequel-4.23.0/lib/sequel/adapters/jdbc.rb:281
    execute_insert at /home/deverly/my_project/vendor/cache/jruby/2.2.0/gems/sequel-4.23.0/lib/sequel/dataset/actions.rb:937
            insert at /home/deverly/my_project/vendor/cache/jruby/2.2.0/gems/sequel-4.23.0/lib/sequel/dataset/actions.rb:336
```

DougEverly:
[...]

PR #1019, opened 2015-06-11
Change to getLong() to accomodate BigInt primary key

DougEverly:
See issue #1018.


jeremyevans:
Thanks for the patch.  This change looks good.  I'll test on Monday, and assuming no problems, it will be merged then.

Other jdbc subadapters will should be checked to see if they need similar changes.  If you want to push commits for other jdbc subadapters, that would be great, but if not I can take care of that after merging this.

DougEverly:
I updated the other JDBC adapters likewise.

jeremyevans:
Thank you.  I'll test all of these on Monday.  Thanks for the patch!

PR #1024, opened 2015-06-20
Adding pg_inet_ops extension

celsworth:
As discussed on IRC.

This isn't finished, as mentioned there are some bizarre spec errors I can't work out, need your help :)

A few points to work out for your thoughts..

* I'm not sure if all the operators and functions I defined are worth having, some of them are just overrides of standard operators (greater_than etc)
* Casting seems to be an issue in the PostgreSQL tests; when using a real column the server works it out, but when passing two strings, they need to be cast to inet and I'm not sure how to achieve this.
* The PostgreSQL tests sometimes return an IPAddr, sometimes a string, I'm not sure if this is consistent? In particular shouldn't the netmask function just return a string perhaps?

Let me know if I can do any more on this, or alternatively feel free to polish it up for inclusion, hope its useful :)

### DIFF
--- a/lib/sequel/extensions/pg_inet_ops.rb
+++ b/lib/sequel/extensions/pg_inet_ops.rb
@@ -1,0 +1,149 @@
+# The pg_inet_ops extension adds support to Sequel's DSL to make
+# it easier to call PostgreSQL inet functions and operators.
+#
+# To load the extension:
+#
+#   Sequel.extension :pg_inet_ops
+#
+# The most common usage is passing an expression to Sequel.pg_inet_op:
+#
+#   r = Sequel.pg_inet_op(:inet)
+#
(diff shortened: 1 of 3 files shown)


jeremyevans:
This looks like a really good start, thanks!  I'll try to fix the remaining issues starting on Monday.  I'll probably remove the methods that are already handled by the existing operators.  I would like to handle the casting issue without having separate classes for inet and cidr, as that seems like overkill.  Maybe a second argument for `pg_inet_op` that defaults to `inet`.

jeremyevans:
About the return values for the PostgreSQL functions, they are expected to return different types: http://www.postgresql.org/docs/9.4/static/functions-net.html.  You'll see there that `netmask` is expected to return an `inet` type.

PR #1026, opened 2015-06-22
Adding option hashes to auto_validations plugin

celsworth:


jeremyevans:
Thanks for the patch. This looks good, the only change I would like to make to it is that instead of the 5 class attr_accessors, I would use a single class attr_accessor named :auto_validation_options, which is a hash containing those 5 hashes.

celsworth:
Alright, I'll make that change.

PR #1028, opened 2015-06-23
Support passing a callable object for DB#listen's timeout parameter.

celsworth:


jeremyevans:
Thanks for the patch! This looks good, I'll test and merge soon.

PR #1029, opened 2015-06-26
Add :columns option to update_refresh plugin

celsworth:


jeremyevans:
Thanks for the patch.  This looks good, I'll merge and test on Monday.

PR #1039, opened 2015-07-11
Make the dirty plugin notice when a serialized column is changed.

celsworth:
This PR makes the dirty plugin a bit more aware of changes happening inside serialized columns.

Previously, changing a serialized column made an entry in `changed_columns`, but not in `column_changes` (and hence `initial_values` etc) which was a bit inconsistent.

Now, everything in the dirty plugin is aware of the change. However this does bring about a new caveat - changing a serialized column from its initial value, then back again, leaves an entry in the dirty plugin structures (which the dirty plugin is supposed to prevent in the first place).

I couldn't see how to fix this (reset_column didn't seem to work) but I still think the new behaviour is an improvement, so I just added a note to the docs about it.


jeremyevans:
This looks good to me.  I'll merge and test on Tuesday.  Thanks for the patch!

PR #1040, opened 2015-07-12
Fix unquoted collate argument when creating columns

celsworth:
I was getting an error trying to use `collate` on PostgreSQL:

```ruby
DB.create_table :configurations do
  primary_key :id
  String :test, null: false, collate: 'C'
end
```
resulted in:

```ruby
Sequel::DatabaseError: PG::UndefinedObject: ERROR:  collation "c" for encoding "UTF8" does not exist
LINE 1: ...figurations" ("id" serial PRIMARY KEY, "test" text COLLATE C ...
```

(unquoted argument to COLLATE)

This PR adds `quote_identifier` to it which fixes it (not sure if it should be `quote_identifier` or `literal` though)

### DIFF
--- a/lib/sequel/database/schema_methods.rb
+++ b/lib/sequel/database/schema_methods.rb
@@ -524,7 +524,7 @@ module Sequel
 
     # Add collate SQL fragment to column creation SQL.
     def column_definition_collate_sql(sql, column)
-      sql << " COLLATE #{column[:collate]}" if column[:collate]
+      sql << " COLLATE #{quote_identifier(column[:collate])}" if column[:collate]
     end
 
     # Add default SQL fragment to column creation SQL.


jeremyevans:
The problem is that the collation is an identifier, so if you don't quote it, it folds to lowercase (on PostgreSQL/MySQL/SQLite), which is why PostgreSQL complains about collation `"c"`.

The issue here is backwards compatibility.  Since the behavior since introduction over 4 years is to use the argument as a literal string, changing things to use quote_identifier will break existing code.  I don't see an easy way to implement this change in a backwards compatible manner, and backwards incompatible changes to the schema methods are not allowed, as they can break migrations.

The best way to implement this change would be to add it via a database extension.  That way backwards compatibility is preserved.

PR #1043, opened 2015-07-21
Add set_extra and update_extra model instance methods

rjclardy:
This PR adds a couple of methods that give you the option of extending your models' allowed columns while performing a set/update instead of overriding them.

### DIFF
--- a/doc/mass_assignment.rdoc
+++ b/doc/mass_assignment.rdoc
@@ -36,7 +36,13 @@ By default, if an invalid setter method call is attempted, Sequel raises a <tt>S
   # Instance level
   post.strict_param_setting = false
 
-In addition to +set_only+ and +update_only+, Sequel also has +set_fields+ and +update_fields+ methods, and these may be a better mass assignment choice for most users.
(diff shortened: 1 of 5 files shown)


jeremyevans:
Thanks for submitting a patch.  I appreciate that you updated the documentation and added the appropriate tests.  Unfortunately, I do not think it makes sense to include this in the base model support, since I don't think the use case is large enough.  For things like this I recommend the use of `Model#set_fields`.

It's possible this could be included as a plugin that ships with Sequel, if you can get support from the community.  You should try asking on the sequel-talk Google Group and see if other users would like this shipped as one of Sequel's plugins.  Alternatively, you can add it as an external plugin, and send me a link to the website for the plugin, and I'll add that link to Sequel's website.

rjclardy:
Okay, I might look into writing a plugin later. My problem with `set_fields` is that it's repetitive. I already defined my `allowed_columns` on the model, so why should I have to redefine them on my set/update calls? That's what makes `set_extra` so useful IMO: it lets me extend my default allowed columns instead of repeating them. I've been monkey patching this recently and rarely use `(set|update)_(fields|only)` anymore.

I'll post to the Google Group as you suggested though and see what other people think. Thanks!

PR #1049, opened 2015-07-24
Execute DB#schema with case sensitive MSSQL 2012

knut2:
The method call `DB#schema(table)` ends with `TinyTds::Error: Invalid object name 'SYS.COLUMNS'.` for a a case sensitive SQL Server DB.
Similar problems happens wit DB#dump_schema_migration.

This changes avoid this problem.

Details see https://groups.google.com/forum/#!topic/sequel-talk/H1l2wB2874M


jeremyevans:
This looks pretty good.  I'll merge locally and test, and assuming no problems this will be pushed to GitHub today.

jeremyevans:
I committed a squashed version of this pull request as 2e97bfe6f8fd5bfa697edd1675ccc9dbffeb55d1

PR #1050, opened 2015-07-26
Add the =~ and !~ operators to ComplexExpression

janko-m:
As discussed on the mailing list, this pull request adds the `=~` and `!~` methods to ComplexExpressions, so now the following query works.

```rb
DB[:movies].where{date.extract(:year) =~ 2015}
```


jeremyevans:
Thanks for the patch!

PR #1054, opened 2015-07-28
Refactor DB2 REORG

karlhe:
As discussed in https://groups.google.com/forum/#!topic/sequel-talk/OwaKtriHeTk

1. Moves code to REORG during `alter_table` and `table_exists?` to DB2 shared adapter.
2. Overrides `apply_alter_table` to REORG after each ALTER TABLE operation.


jeremyevans:
Thanks for the patch!  This looks good to me.  I'll test it on my virtual machine with DB2, and will merge it if there are no errors.

PR #1055, opened 2015-07-30
add support for loading lazy columns as a group

brianphillips:
This helps with class table inheritance where we use lazy loading to
load columns from the sub-class table(s). With this change, we load all
the sub-class columns at once instead of retrieving one column at a
time.


jeremyevans:
As discussed on IRC, I'll close this as we're going to use a different API for it.

PR #1060, opened 2015-08-05
Save a reference to the dataset with Sequel::NoMatchingRow

pedro:
This way error handlers can inspect the exception to get the table name or even the model that originated it (if any).


jeremyevans:
I think this makes sense to add to Sequel.  The code looks good, I'll try merging and testing right away.  Thanks for the patch!

pedro:
Thank you!

PR #1065, opened 2015-08-12
Clarify :return option, it was a little ambiguous.

celsworth:
(having the :value made it look like it should be a hash inside
the :return value, which isn't right)


jeremyevans:

PR #1067, opened 2015-08-14
Fix indentation in rdoc for self.register

celsworth:
It wasn't parsing properly: http://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Postgres/PGRange.html#method-c-register


jeremyevans:

PR #1069, opened 2015-08-17
rewrite resources list in readme

Larivact:
You shouldn't have to hover over every item in the resources list you should be able to just read it.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -17,12 +17,12 @@ toolkit for Ruby.
 
 == Resources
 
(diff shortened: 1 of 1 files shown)


jeremyevans:
I don't understand why you are treating the GitHub Issues, Google Group, and IRC links differently than the other links.  But I'm fine with the change to display the links.  I'll merge this and refactor it.  Thanks for the patch!

Larivact:
looks good :+1:

PR #1078, opened 2015-09-17
Strips non-numeric chars w/ typecasting BigDecimal

presidentJFK:


presidentJFK:
I wasn't sure if gsub was going too be slow for this case, but could not think of a reasonably simple alternative.

jeremyevans:
Thanks for submitting a patch, especially one that includes tests.  However, I don't think we want this as a default, as Sequel attempts to be strict by default. For loose typecasting, Sequel ships with the looser_typecasting Database extension.  Note that that extension handles things differently, since `1,0` is treated as `1.0`, not `10`.  If you do want `1,0` treated as `10`, you should just add your own extension that overrides `typecast_value_decimal`.

presidentJFK:
No problem, thanks for getting back to me fast! I'll work on an extension.

PR #1083, opened 2015-09-29
Transaction Synchronization on MRI

YorickPeterse:
This PR fixes two things:

1. Synchronization on MRI when calling `Sequel.synchronize`
2. Synchronization of transaction options, without the need for global locks

Both commits explain the problem and rationale pretty well, so see those for more information. If anything has to be changed styling wise, just say the word.

### DIFF
--- a/lib/sequel/core.rb
+++ b/lib/sequel/core.rb
@@ -388,7 +388,7 @@ module Sequel
 
   private_class_method :adapter_method, :def_adapter_method
 
-  require(%w"deprecated sql connection_pool exceptions dataset database timezones ast_transformer version")
+  require(%w"deprecated sql connection_pool exceptions dataset database timezones ast_transformer version synchronized_hash")
 
   # Add the database adapter class methods to Sequel via metaprogramming
   def_adapter_method(*Database::ADAPTERS)
--- a/lib/sequel/core.rb
+++ b/lib/sequel/core.rb
@@ -293,24 +293,14 @@ module Sequel
     end
   end
 
-  if defined?(RUBY_ENGINE) && RUBY_ENGINE != 'ruby'
-  # :nocov:
-    # Mutex used to protect mutable data structures
-    @data_mutex = Mutex.new
+  # Mutex used to protect mutable data structures
+  @data_mutex = Mutex.new
 
-    # Unless in single threaded mode, protects access to any mutable
-    # global data structure in Sequel.
-    # Uses a non-reentrant mutex, so calling code should be careful.
-    def self.synchronize(&block)
-      @single_threaded ? yield : @data_mutex.synchronize(&block)
-    end
-  # :nocov:
-  else
-    # Yield directly to the block.  You don't need to synchronize
-    # access on MRI because the GVL makes certain methods atomic.
-    def self.synchronize
-      yield
-    end
+  # Unless in single threaded mode, protects access to any mutable
+  # global data structure in Sequel.
+  # Uses a non-reentrant mutex, so calling code should be careful.
+  def self.synchronize(&block)
+    @single_threaded ? yield : @data_mutex.synchronize(&block)
   end
 
   # Uses a transaction on all given databases with the given options. This:
--- a/lib/sequel/database/misc.rb
+++ b/lib/sequel/database/misc.rb
@@ -125,7 +125,7 @@ module Sequel
       @schemas = {}
       @default_string_column_size = @opts[:default_string_column_size] || DEFAULT_STRING_COLUMN_SIZE
       @prepared_statements = {}
-      @transactions = {}
+      @transactions = SynchronizedHash.new
       @identifier_input_method = nil
       @identifier_output_method = nil
       @quote_identifiers = nil
--- a/lib/sequel/database/transactions.rb
+++ b/lib/sequel/database/transactions.rb
@@ -159,10 +159,9 @@ module Sequel
       end
     end
 
-    # Synchronize access to the current transactions, returning the hash
-    # of options for the current transaction (if any)
+    # Returns the hash of options for the current transaction (if any)
     def _trans(conn)
-      Sequel.synchronize{@transactions[conn]}
+      @transactions[conn]
     end
 
     # Add the current thread to the list of active transactions
--- a/lib/sequel/synchronized_hash.rb
+++ b/lib/sequel/synchronized_hash.rb
@@ -1,0 +1,35 @@
+module Sequel
+  # A Hash with synchronized access. This class only provides methods actually
+  # used by Sequel (e.g. [] and []=).
+  class SynchronizedHash
+    def initialize
+      @mutex = Mutex.new
+      @hash  = {}
+    end
+
+    # Returns the value of the given key
+    def [](key)
+      @mutex.synchronize { @hash[key] }
+    end
+
+    # Sets the key +key+ to +value+
+    def []=(key, value)
+      @mutex.synchronize { @hash[key] = value }
+    end
+
+    # Deletes the given key
+    def delete(key)
+      @mutex.synchronize { @hash.delete(key) }
+    end
+
+    # Returns the keys of the Hash
+    def keys
+      @mutex.synchronize { @hash.keys }
+    end
+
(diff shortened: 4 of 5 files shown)


YorickPeterse:
There are probably more cases where at least the global lock (via `Sequel.synchronize`) isn't needed, or perhaps other places where synchronization isn't entirely done properly. Either way, this PR only focuses on transaction handling as the code in question has been throwing a bunch of errors (and possibly segmentation faults) in production.

jeremyevans:
First, I want to thank you for your extensive work on this issue.  Coming up with reproducible examples for thread-safety issues can be very difficult due to non-determinism, and your commit messages were very descriptive.

After additional discussion on IRC, I'm not positive there are issues in Sequel here, though there certainly may be.  I've been running the example crashing program for almost 90 minutes with no crashes, but I may just be getting lucky.

Assuming that the segfaults you are hitting are due to thread safety issues in Sequel, I'm not sure these patches are the best way to handle it.  The unstubbing of `Sequel.synchronize` I'm sort of OK with.  It shouldn't be necessary on MRI, and there would be a small speed hit in unstubbing.  The reason it shouldn't be necessary is `Sequel.synchronize` should only be used for data structure access, not for generic synchronization.  The basic use case is `Sequel.synchronize{HASH[value]}`, `Sequel.synchronize{HASH[value1] = value2}`, `Sequel.synchronize{HASH.delete(value)}`.  Due to the GIL on current MRI, it shouldn't be necessary to use a mutex for these operations.  If `Sequel.synchronize` is used for something else, the other code should probably be changed to use  a different mutex.  That being said, unstubbing would be simpler and more consistent across interpreters.  I'll do some benchmarking and if the performance hit isn't too bad, I can accept that patch.

The synchronized hash for transaction options I think is a bad idea.  Because of the default connection pool, two threads should never have access to the same connection object at the same time, so it definitely shouldn't be necessary.  If two threads are getting access to the same connection at the same time using the default connection pool, the bug should be fixed there, we shouldn't be trying to work around the bug in the transaction code.

The gdb trace provided (https://gist.github.com/YorickPeterse/4c89539c3eb509469b69), shows only a single thread in ruby land at the time of the crash, and it appears to be due to memory corruption of the the underlying hash instance, as oppose to simultaneous mutation by another thread.  It's possible that the memory corruption was due to earlier thread safety problems before the crash, though.  However, assuming that this is the same error given by the ruby segfault output (https://gist.github.com/YorickPeterse/8d3cd78488cfed4e134a), in this case we are taking about the `@transactions` hash, not the hash returned by `@transactions[conn]`, so it doesn't look like the synchronized hash patch would affect that issue.

If other people could run the example crashing program (https://gist.github.com/YorickPeterse/188b540f13ee0de538e8) and provide feedback (crashed within x minutes, didn't crash after x minutes, ruby output, gdb output) with details on their environment (operating system version, ruby version, pg/jdbc-postgresql version, compiler used, and flags ruby was compiled with), that would be very helpful.

janko-m:
It didn't crash after 54 minutes on MRI.

Mac OS Yosemite
`ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14]`
`pg 1.8.3`
`'--prefix=/Users/janko/.rbenv/versions/2.2.2' '--with-openssl-dir=/usr/local/opt/openssl' '--with-readline-dir=/usr/local/opt/readline' 'CC=clang' 'CFLAGS= -O3 -Wno-error=shorten-64-to-32 ' 'LDFLAGS=-L/Users/janko/.rbenv/versions/2.2.2/lib ' 'CPPFLAGS=-I/Users/janko/.rbenv/versions/2.2.2/include '`

jeremyevans:
Using the example crashing program and some helpful testers, I've got a confirmed crash on Linux 3.16, and it not crashing on Mac OS X, OpenBSD, or Windows. I've also had a report of it not crashing on Linux 3.16 after about 45 minutes with the unstubbing of `Sequel.synchronize` patch provided (the first commit in this pull request, without the second one).  This may be a Linux-only race condition in MRI's threading that Sequel just happens to hit.

Since stubbing `Sequel.synchronize` on MRI is basically taking advantage of an MRI implementation detail, and it no longer appears safe to do that in all cases, I'm going to merge the first commit.  The second commit I don't think is good idea because access to connection objects should already be synchronized by the connection pool, so even if it does fix something, it's would be hiding the true cause of the bug.

YorickPeterse:
@jeremyevans Sounds good to me. I'll do some extra digging tomorrow to see if I can pinpoint exactly what's getting corrupted, where and how (it's currently quite late over here).

The interesting part is that the assignment of transactions options is synchronized (https://github.com/jeremyevans/sequel/blob/3cab4260cdfa5a2f6f4e25476b460861d6d7564b/lib/sequel/database/transactions.rb#L186-L188), so is the code that deletes the transaction options of a connection. 

It's worth mentioning that due to this being timing sensitive it can be _very_ hard to reproduce using the script I provided. Cranking up the amount of threads _might_ help, but it's a hit and miss. For good measurement, I'm using Linux 4.1.6 with pg 0.18.3, Ruby 2.2.2, and PostgreSQL 9.4.4. So far I've only been able to reproduce this on MRI, but I'll give JRuby and Rubinius a try tomorrow. If the issue is indeed due to improper synchronization of some form then there's chance the problem will also manifest on JRuby and/or Rubinius.

YorickPeterse:
From the top of my head, I can think of the following theories causing these problems:

* Sequel's connection pooling has a fault, resulting in connection objects being shared between multiple threads when not expected. Concurrent, unsynchronized modifications (such as those performed by `Sequel::Database#transaction_finished?`) could lead to the underlying Hash getting corrupted.
* Not synchronizing code on MRI was the problem, my first commit fixed this, and I didn't bother to properly check if my 2nd commit actually fixed any thing. I'm pretty sure I did check things, but I know myself well enough by now. Verifying this should be fairly easy.
* Connection objects are synchronized, so is access to `@transactions[conn]`. However, two threads end up concurrently modifying `@transactions`, unrelated to the connection objects they're using.
* I'm full of poop and it's just my environment that's utterly broken. This is unlikely since somebody else reported crashes as well.

jeremyevans:
@YorickPeterse I gave another look at the second commit in this pull request, and I don't think it will help.  There are only three times `@transactions` is used in Sequel:

```
$ fgrep -r @transactions lib
lib/sequel/database/misc.rb:      @transactions = {}
lib/sequel/database/transactions.rb:      Sequel.synchronize{@transactions[conn]}
lib/sequel/database/transactions.rb:        Sequel.synchronize{@transactions[conn] = hash}
lib/sequel/database/transactions.rb:        Sequel.synchronize{@transactions.delete(conn)}
```

The first line here is during Database instance initialization, the other three of these as you can see are already protected by the same mutex.  The second commit only makes `@transactions` a synchronized hash, it doesn't make `@transactions[conn]` entries synchronized hashes.  So I don't think it will help.

In terms of your theories, I'm leaning toward the second theory.  I was eventually able to make Windows crash without the first commit (I think, it didn't give any output, just stopped), but it is still running with the first commit.

The first theory is possible, but there has been a huge amount of thread safety testing of the connection pool code.  It's possible there are still bugs there, but I doubt it.

The third theory of two threads modifying `@transactions` is very unlikely, as shown above.

I fourth theory is wrong as well, as we've been able to reproduce this in certain cases.

I want to thank you again for your work on this.  Your test program was very valuable in pointing out the underlying issue, and it shows that you really cannot rely on the GIL in MRI even for protecting simple data structure access, as least in certain environments.  This is going to cause me to reassess some of my other libraries, as I know at least Roda makes the same assumption as Sequel.

jeremyevans:
OK, I've merged the first commit.  I'm going to close this now.  If there are still thread safety issues after this commit, please let me know so that we can investigate and possibly reopen.

YorickPeterse:
To confirm/rule out the theory about only my first commit making a difference I ran my test again using only that commit (thus excluding the one adding SynchronizedHash). After running this for over an hour I've yet to see it crash. Either the particular problem has been solved, or it has become significantly harder to reproduce. In both cases this should be good enough for now, at least it no longer segfaults after a few minutes. Thanks!

PR #1089, opened 2015-10-11
Silence warnings

aptinio:
Hi Jeremy,

I've read your position regarding warnings on sequel-talk but I'm hoping to show warnings in my projects that depend on `sequel`. What do you think about the strategies used in this PR? Specifically, the use of:
```ruby
instance_variable_defined?(:@foo) && @foo
```
and
```ruby
@foo ||= nil
```

If you're okay with these, I can continue applying these strategies to the rest of the code.
Pardon the white space related changes. I've set my editor to automatically remove trailing spaces. Let me know if you'd like to do away with these.


jeremyevans:
As I've stated repeatedly, I don't think initializing instance variables to nil to avoid VERBOSE warnings is worth it from a cost/benefit perspective.  Setting the instance variables to nil (`@foo ||= nil`) has a significant memory penalty.  Checking for instance variable before every use has a significant performance penalty (3-7 times slower), even when using `defined?` instead of `instance_variable_defined?` (`defined?(@foo)` is faster). As long as ruby VERBOSE warns for uninitialized instance variables and does not offer a zero-cost way to avoid them, I don't plan for Sequel to be VERBOSE warning free.

I'll definitely consider patches to fix other VERBOSE mode warnings other than uninitialized instance variables. VERBOSE has its uses, and Sequel actually has rake tasks that run in VERBOSE mode and filter out the uninitialized instance variable warnings (e.g. `rake spec_core_w`).  For code that uses Sequel that you want to test with VERBOSE, I recommend a similar filtering strategy.

As mentioned in the contribution guidelines, please do not mix whitespace changes with code changes in the future.

aptinio:
Thanks for explaining. I'll look into applying the filtering strategy that you used. I missed the contributing guide, sorry about that.

Issue #67, opened 2011-05-01
Sequel::Model.create_table now (Sequel 0.1.9.12) broken


jeremyevans:
1. Setup a Model with class MyModel < Sequel::Model.
2. Define a schema using Sequel:::Model.set_schema do.
3. Invoke Sequel::Model.create_table

I'd expected that this 'll create a table for my model, but instead I get NoMethodError right into my 
face telling me that Sequel::Schema::Generator#create_sql does not exist.

I'm using Sequel 0.1.9.12...

Google Code Info:
Issue #: 37
Author: florian....@googlemail.com
Created On: 2007-08-28T11:17:34.000Z
Closed On: 2007-08-28T11:47:55.000Z

PR #1091, opened 2015-10-18
note timestamp values always loaded as Time objects

odigity:
(Whether the column is defined using Time or DateTime.)

### DIFF
--- a/doc/schema_modification.rdoc
+++ b/doc/schema_modification.rdoc
@@ -50,7 +50,7 @@ convert to database types:
 
 Note that in addition to the ruby class name, Sequel also pays attention to the column options when
 determining which database type to use.  Also note that for boolean columns, you can use either
-TrueClass or FalseClass, they are treated the same way (ruby doesn't have a Boolean class).
+TrueClass or FalseClass, they are treated the same way (ruby doesn't have a Boolean class). Also note that whether you specify DateTime or Time types, the value will be loaded from the database as a Time object.
 
 Also note that this conversion is only done if you use a supported ruby class name.  In all other
 cases, Sequel uses the type specified verbatim:


odigity:
Caveat:  I'm assuming this is true based on my test run on MySQL:

    create_table(:temptest) do
      String   :name
      DateTime :type_datetime
      Time     :type_time
    end

    tnow  = Time.now
    dtnow = DateTime.now
    temptest = self[:temptest]
    temptest.insert(name: 'matching',    type_datetime: dtnow, type_time: tnow)
    temptest.insert(name: 'mismatching', type_datetime: tnow,  type_time: dtnow)

---

    MariaDB [apps]> describe temptest;
    +---------------+--------------+------+-----+---------+-------+
    | Field         | Type         | Null | Key | Default | Extra |
    +---------------+--------------+------+-----+---------+-------+
    | name          | varchar(255) | YES  |     | NULL    |       |
    | type_datetime | datetime     | YES  |     | NULL    |       |
    | type_time     | datetime     | YES  |     | NULL    |       |
    +---------------+--------------+------+-----+---------+-------+
    3 rows in set (0.00 sec)

---

    irb(main):007:0> db[:temptest].map { |r| puts "name: #{r[:name]}, dt: #{r[:type_datetime]} (#{r[:type_datetime].class}) t: #{r[:type_time]} (#{r[:type_time].class})" } 
    I, [2015-10-18T13:37:35.658358 #67]  INFO -- : (0.000176s) SELECT * FROM `temptest`
    name: matching, dt: 2015-10-18 13:34:50 +0000 (Time) t: 2015-10-18 13:34:50 +0000 (Time)
    name: mismatching, dt: 2015-10-18 13:34:50 +0000 (Time) t: 2015-10-18 13:34:50 +0000 (Time)

odigity:
Forgot to check one more thing, and I found a surprise there when I did -- the post-migration schema dump contains this:

    create_table(:temptest) do
      String :name, :size=>255
      DateTime :type_datetime
      DateTime :type_time
    end

Which is odd, given that Sequel is normalizing date/time values to Time objects when loading them from the database...

jeremyevans:
In general, the ruby type used for database timestamp types depends on the `Sequel.datetime_class` setting, which is `Time` by default but can be set to `DateTime`.  Technically, it depends on the adapter/driver, some of which do not respect `Sequel.datetime_class`.  So I have to reject this as it is not completely accurate.  I'm open to a new or revised pull request that provides more accurate information, though.

PR #1092, opened 2015-10-26
Fix broken link in validations.rdoc

celsworth:
This AR guide seems to have been split into two (Callbacks and Validations) since the link was made.


jeremyevans:
Thanks for the patch!

PR #1094, opened 2015-10-26
Update docs to reflect transaction use in cursors

adarsh:
Reason for change:
* The documentation wasn't clear about the default transaction when using Postgres cursors and I found myself stuck.
* The nice `use_cursor` method takes an option of `hold: true` which prevents rolling back all changes if a block fails, for example iterating over 10k records.

What the change was:
* Clarify that `use_cursor` uses a transaction by default and how to disable it.
* Add cursors to the Transactions document as another method which uses them by default.

Relevant discussion on the mailing list: https://groups.google.com/forum/#!topic/sequel-talk/7IRBCE-Rwv0

### DIFF
--- a/doc/transactions.rdoc
+++ b/doc/transactions.rdoc
@@ -6,6 +6,7 @@ Sequel uses autocommit mode by default for all of its database adapters, so in g
 * Model#save
 * Model#destroy
 * Migrations if the database supports transactional schema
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch! I'll get this merged shortly.

PR #1095, opened 2015-10-29
fix `class java.lang.String` error when jdbc adapter with logstash

eriktrom:
fix error `cannot convert instance of class org.jruby.RubyObject to class java.lang.String`
when providing a `jdbc_password` input field to the logstash-input-jdbc jruby plugin

reasoning: https://github.com/elastic/logstash/issues/3785#issuecomment-152104199

### DIFF
--- a/lib/sequel/adapters/jdbc.rb
+++ b/lib/sequel/adapters/jdbc.rb
@@ -213,7 +213,7 @@ module Sequel
             props = java.util.Properties.new
             if opts && opts[:user] && opts[:password]
               props.setProperty("user", opts[:user])
-              props.setProperty("password", opts[:password])
+              props.setProperty("password", opts[:password].value)
             end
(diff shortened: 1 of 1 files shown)


eriktrom:
@jeremyevans - I totally realize this is likely the wrong place for this issue, that said, I forked to fix this issue and left a comment in a hardly related thread where another user 3 days ago had the same issue. 

https://github.com/elastic/logstash/issues/3785#issuecomment-152104199

I'm not sure what else to do except learn the ecosystem just find the root cause, but that's not likely this evening, so figured I'd share and link. Perhaps you know more, if not, just close this. It's a ditch effort to help with something I know nothing about. :) 

cheerio

eriktrom:
scratch that, found it: https://github.com/eriktrom/logstash-input-jdbc/commit/adda69e5eba68d32907e58aad0af9af656b3d2a0 

closing, sorry

PR #1101, opened 2015-11-12
Fix the :rank option to Dataset#full_text_search on Postgres to order…

chanks:
… by rank descending.

Postgres' `ts_rank_cd` function returns a rank where a higher number represents a closer association. Currently the `:rank` option to `full_text_search` orders by `ts_rank_cd(...)` rather than `ts_rank_cd(...) DESC`, which means users will get the least relevant search results first. This is probably not what anyone is expecting :smile:

### DIFF
--- a/lib/sequel/adapters/shared/postgres.rb
+++ b/lib/sequel/adapters/shared/postgres.rb
@@ -1331,7 +1331,7 @@ module Sequel
         end
 
         if opts[:rank]
-          ds = ds.order{ts_rank_cd(cols, terms)}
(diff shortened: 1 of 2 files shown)


chanks:
Also, there wasn't any guidance in the "contributing" doc on whether to include a changelog entry, so I didn't - let me know if you want me to add one.

jeremyevans:
Good catch.  This makes sense to me, I'll merge it shortly.

In general I take care of adding CHANGELOG entries.  I don't have a problem if someone adds them in their pull request, but it makes merge conflicts more likely, so I don't encourage it.

PR #1105, opened 2015-11-17
Add a Sequel::Model#was_new? method

glittershark:
Add a `Sequel::Model#was_new?` method, that returns true if the model
instance was just created, as opposed to having been fetched from the
database.

fwiw, my current use case is a helper method that returns HTTP 201 if 
a model was just created, and 200 if it was updated

### DIFF
--- a/lib/sequel/model/base.rb
+++ b/lib/sequel/model/base.rb
@@ -117,6 +117,7 @@ module Sequel
       def call(values)
         o = allocate
         o.instance_variable_set(:@values, values)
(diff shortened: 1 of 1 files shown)


jeremyevans:
Sorry, but I don't think it makes sense to make this the default behavior.  For one, it's a performance hit to the retrieval of all model objects.

It should be fairly easy to add this feature via a plugin, I suggest you try that out and see how it works.  If you'd like such a plugin to ship with Sequel, please open up a thread on the sequel-talk Google Group to see if there is any community support for the idea.

PR #1106, opened 2015-11-18
Add a UUID plugin.

pdrakeweb:


jeremyevans:
This plugin looks good, thanks for the patch!  I know other people have asked for similar functionality in the past, so I think there is a need for this, and a plugin is the right way to address it.  I'll try to test and merge this later today.

PR #1108, opened 2015-11-27
Add sqlite3 dependency to examples

olivierlacan:
Using the code in the examples yielded an exception because sqlite3 was not required. 
It seems like that dependency should be shown in the examples.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -43,6 +43,7 @@ If you have any comments or suggestions please post to the Google group.
 == A Short Example
 
   require 'sequel'
+  require 'sqlite3'
   
   DB = Sequel.sqlite # memory database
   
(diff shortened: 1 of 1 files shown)


janko-m:
Hmm, I was sure that it's enough to have it in the Gemfile, but I'm not at my computer to check. We'll see what Jeremy says.

jeremyevans:
Sequel automatically requires sqlite3 if you use `Sequel.sqlite` or `Sequel.connect('sqlite:/')`, so this shouldn't be necessary.  It is true the current code will fail if sqlite3 is not installed, but the same is true of the code after this change.

I do agree that we should probably mention somewhere that sqlite3 is required.  There are also examples in the README where the postgres adapter is used, and we should mention that pg is required there.  I'll fix this shortly.

jeremyevans:
Fixed by 40d74e9cafc1aae4ed6ce91a306e1c9dccbbfea7

PR #1113, opened 2015-12-09
allow IP6 hosts in database URL

hellvinz:
see
https://github.com/rails/rails/commit/1c2f52175008ffd4c31338138f5b11bff677123b


jeremyevans:
Thanks for the patch!  This makes sense to me, as long as the tests pass it will be merged shortly.

hellvinz:
thanks for your quick answer
I've added some tests but I'm not sure I've put them in the right file.

jeremyevans:
Looks like this breaks on Ruby 1.8.7, which Sequel still supports.  I'll modify things to make it conditional.

PR #1115, opened 2015-12-11
Support :eager_limit and :eager_limit_strategy Dataset options.

chanks:
As discussed on sequel-talk. A couple of notes:

- The eager-loading association logic is a bit messier, by virtue of needing to passing around limit and offset information between methods, rather than being able to simply pull it from the association reflection when it was necessary. For backwards compatibility I made sure to not change the arguments required by any methods.
- I've added the support for Dataset#eager, but not Dataset#eager_graph. I started on the work for #eager_graph, but it seemed like it was going to be a more complicated addition and I wasn't planning on actually using the feature with #eager_graph, so I figured it would be safer and simpler to add it for #eager only and just document that the feature only works there. If you think it's important to also support it for #eager_graph so that the APIs are similar, I can certainly add it. Or, if you think it's fine to only support #eager with this feature, I can add documentation to that effect.

### DIFF
--- a/doc/advanced_associations.rdoc
+++ b/doc/advanced_associations.rdoc
@@ -287,15 +287,35 @@ the window function strategy:
     :eager_limit_strategy=>:window_function
   Artist.where(:id=>[1,2]).eager(:first_10_albums).all
   # SELECT * FROM (
-  #   SELECT *, row_number() OVER (PARTITION BY tracks.album_id ORDER BY release_date) AS x_sequel_row_number_x
-  #   FROM tracks
-  #   WHERE (tracks.album_id IN (1, 2))
+  #   SELECT *, row_number() OVER (PARTITION BY albums.artist_id ORDER BY release_date) AS x_sequel_row_number_x
+  #   FROM albums
+  #   WHERE (albums.artist_id IN (1, 2))
   # ) AS t1
   # WHERE (x_sequel_row_number_x <= 10)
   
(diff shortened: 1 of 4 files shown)


jeremyevans:
Thanks for the patch!  This seems like the minimally invasive approach to add the functionality.  I'm fine with this only working with eager and not eager_graph for now, as long as that behavior is documented.  Could you add documentation on this to the advanced_associations guide, including an example of usage? Also, could you add an integration test for this feature, showing it works correctly with different limits passed in?

chanks:
No problem!

chanks:
Added an integration spec and a section in the advanced_association doc.

One awkward part of the :eager_limit implementation is that it doesn't seem to interact well with the :union strategy, since that strategy is explicitly rejected if there's an :eager_block present. I'm not certain why that is, but it's a little annoying that the :union strategy might be best for a given query, but the usage of :eager_limit, even if it's the only modification made in the block, makes it unusable. I'm not sure what the solution is for that.

jeremyevans:
The union strategy is the best strategy in general, which is why it is the default strategy.  Unfortunately, the way it is implemented caches the union eager loader in the association, so it cannot be used for per-call limits.  It is possible that you could get the union eager strategy working, but it would require more work, and the union strategy requires a lot more SQL creation, which may offset the performance benefits if it can't be cached.

However, I don't think supporting the union strategy is required for merging.  Just the flexibility of per-call limits is an important addition.

PR #1116, opened 2015-12-11
Fix typo in Dataset#insert documentation

wallace:
"an" is changed to "and"


jeremyevans:

PR #1121, opened 2015-12-28
Add Dataset#insert_conflict on SQLite

Sharpie:
SQLite has supported alternate conflict resolution algorithms since version
2.3.0. This patch adds support for `insert_conflict` which allows users to
specify an alternate algorithm such as IGNORE or REPLACE. A `insert_ignore`
method is also added for compatibility with MySQL.


jeremyevans:
This looks good to me, thanks for the patch!  Assuming this passes tests, I'll merge it.

PR #1122, opened 2016-01-03
Update copyright notice to 2016 [ci skip]

Atul9:


jeremyevans:
I haven't written any code for Sequel in 2016 yet, so updating the copyright notice is premature.  Also, only the author should be updating the copyright notice. This change should be made by me on the first significant commit of the year.

Issue #1139, opened 2016-02-04
Memory leak when not using connection pooling


perlun:
Hi,

We've just concluded a debugging session for one of our apps, which are running in production for a few years. The application server (using JRuby 1.7) tends to run out of memory over time, so the customer has scheduled reboots of the machine in place, as a workaround.

Today, we debugged this a bit more and found out the following _very_ interesting facts:

- If connection pooling is enabled, all is fine; the memory usage over time does not increase drastically. After 100 000 connect/disconnect cycles, the memory usage for the Java process was still only around 30 MiB which is completely fine.
- If connection pooling is _disabled_ (as in the script below), the process leaks a huge amount of memory. 100 000 connect/disconnect cycles brought it up to around 750 MiB on my machine. Note: the is specifically tested with JRuby, with the AS400 driver (which is the one used in production also). We tested with the Postgres driver first, which also seemed to leak memory (and we didn't investigate that so carefully, since the "real" use case is the AS400 driver for us).

Here are the results when using unpooled connections (the full script below):

![image](https://cloud.githubusercontent.com/assets/630613/12814964/cae05628-cb4b-11e5-9628-87e81be0dd45.png)

...and when using the `connect_pooled` method instead:

![image](https://cloud.githubusercontent.com/assets/630613/12814982/f23a192a-cb4b-11e5-9d3f-d90eebb84219.png)

----

Running Sequel 4.31.0, JRuby 1.7.22, Java 1.7.0_80.

```ruby
#!/usr/bin/env java -XX:+UseConcMarkSweepGC -XX:MaxPermSize=128m -XX:+CMSClassUnloadingEnabled -Djruby.jit.max=-1 -Djruby.thread.pooling=true -Djjruby.compile.mode=FORCE -Dfile.encoding=UTF-8 -Djava.net.preferIPv4Stack=true -jar jruby-complete-1.7.22.jar

require 'pry'
require 'sequel'

require 'java'
require './jt400.jar'

def connection_string
  'jdbc:as400://127.0.0.1;prompt=false;user=foo;password=bar;libraries=baz;query timeout mechanism=cancel;'
end

def connect
  Sequel.connect(connection_string).tap { |db|
    db.identifier_output_method = nil
    db.identifier_input_method = nil
  }
end

def connect_pooled
  @connection ||= Sequel.connect(connection_string, max_connections: 10, pool_timeout: 60).tap { |db|
    db.extension(:connection_validator)
    db.identifier_output_method = nil
[...]

PR #1140, opened 2016-02-04
Added notes about how to avoid memory leaks

perlun:
This is a basically a followup to https://github.com/jeremyevans/sequel/issues/1139, trying to help other people avoid making the same mistake that I have been doing. :smile:


jeremyevans:
Looks good, thanks!

PR #1143, opened 2016-02-10
JDBC: #schema_parse_table to include remarks

olleolleolle:
Comments on columns are metadata in the schema. We can include them in the schema metadata.

[DatabaseMetaData.html#getColumns](https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getColumns(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String))


jeremyevans:
Thanks for submitting a pull request. We don't include comments in the schema parsing of  any of the other adapters, but I suppose there is no harm in including them here considering JDBC has already done the work for us.  I'll merge and test later today.

PR #1149, opened 2016-02-16
Fix a bad "it's."

nirvdrum:


jeremyevans:

PR #1155, opened 2016-03-13
Add validate_numericality_of validator

petedmarsh:
This that checks if a value is less/greater/equal than a specified
value, e.g.:

    class Example < Sequel::Model
        validates_numericality_of :something, less_than: 123
    end

The validator supports the following options (which may be used in
conjunction):

    :equal_to
    :less_than
    :less_than_or_equal_to
    :greater_than
    :greater_than_or_equal_to

NOTE: This is based on ActiveRecord's validate_numericality_of
validator, code:

    https://github.com/rails/rails/blob/2fc6c094e54103db2a4550af6e2ac2f34420f084/activemodel/lib/active_model/validations/numericality.rb

As both ActiveRecord and Sequel are MIT licensed I believe this is
fine.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,8 @@
 
 * Allow foreign_key schema method to handle SQL::Identifier and SQL::QualifiedIdentifier as 2nd argument (jeremyevans)
 
+* Add validate_numericality_of validator for less/greater/equal validations (petedmarsh)
+
 === 4.32.0 (2016-03-01)
 
 * Use mutex for synchronizing access to association reflection cache on MRI (jeremyevans)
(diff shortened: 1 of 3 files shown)


jeremyevans:
I think the following API would be better, as it is more similar to how Sequel handles other validations:

```ruby
validates_equal(5, :something)
validates_less_than(5, :something)
validates_greater_than(5, :something)
validates_not_less_than(5, :something)
validates_not_greater_than(5, :something)
```

For one, this API isn't specific to numbers, it could be used with any type implementing Comparable (and `validates_equal` would work for all types).

That being said, I'm not sure these validations are common enough to warrant inclusion in the validation_helpers plugin.  Could you open up a thread on the sequel-talk Google Group to see what other users think about these validations?  Depending on the reception there, I'll make a decision about whether to include these validations.

Closing now as this API will not be accepted.

petedmarsh:
I like your proposed API - I've created a thread in sequel-talk (which I think may be in a review queue or similar as I can't see it in the group yet). If there's a consensus that these would be useful I'll be happy to re-implement with the API proposed above :)

jeremyevans:
Your message should be public in the group now.  Thanks for opening up a thread, we'll see what the community thinks.

PR #1158, opened 2016-03-20
Add "none" section to active_record docs

NickTomlin:
This adds a tip for users wishing to replicate AR's `none`. The example is taken from from [this](http://stackoverflow.com/a/36087167/1048479) SO answer.

Points of interest:

1. I'm not super famliar with RDOC, please let me know if I've made gaff
2. I'm not sure if the example should encourage extending every dataset. It seems like the clearest and cleanest way to do so (and fitting with how AR handles things)

### DIFF
--- a/doc/active_record.rdoc
+++ b/doc/active_record.rdoc
@@ -626,6 +626,22 @@ For anything more complex, you can use +dataset_module+:
     end
   end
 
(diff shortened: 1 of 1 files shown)


jeremyevans:
Looks good, thanks!

PR #1159, opened 2016-03-22
Adding :find option for nested_attributes method from NestedAttributes extension

AlexWayfer:
...with tests for these cases

### DIFF
--- a/lib/sequel/plugins/nested_attributes.rb
+++ b/lib/sequel/plugins/nested_attributes.rb
@@ -107,6 +107,7 @@ module Sequel
         #                  object. Set to :create to create a new object with that primary
         #                  key, :ignore to ignore the record, or :raise to raise an error.
         #                  The default is :raise.
+        # :find :: Allow find existing nested records instead creating
         #
         # If a block is provided, it is used to set the :reject_if option.
         def nested_attributes(*associations, &block)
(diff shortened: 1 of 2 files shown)


lib/sequel/plugins/nested_attributes.rb
```diff
@@ -171,6 +172,30 @@ def nested_attributes_create(meta, attributes)
...
+          nested_attributes_add_association_object(reflection, obj)
+        end
+
+        # Find object with the given attributes, and add it to associations
+        # Returns the object found.
+        def nested_attributes_find(meta, attributes)
+          reflection = meta[:reflection]
+          obj = reflection.associated_class.find(attributes)
+          # Additional compliance checking of attributes
+          return if !obj || attributes.any? { |k, v| obj.values[k] != v }
```
AlexWayfer: It works wrong without `attributes.any?`, look for [this subject](https://groups.google.com/forum/#!topic/sequel-talk/R1dt7MSbKr0).

jeremyevans:
I may not be understanding the reasoning behind this.  It looks like this change is making a query to the database using all fields in the table, and then saving all fields with the same values.  It looks like what you want is to just associate objects to the main record based on attribute values, which is not the use case for nested_attributes. nested_attributes is designed to update fields of associated records at the same as the current record. I think you probably should just write some custom code to find objects and associate them, instead of using nested attributes.

I'm going to leave this open for a little while, as I think I may be missing something or I haven't understood this properly.

AlexWayfer:
I want use something like this:
```ruby
album_params = { name: 'The Second Album', artist_attributes: { name: 'Existing artist or not' } }
Album.create(album_params)
```
If artist not exists - it will be created him. If artist exists - it will be associate with him, not create other with same name.
For creating (or finding) many objects by mass params without duplicates.

jeremyevans:
Ah, that makes more sense.  However, I don't think this is a good fit for the nested_attributes plugin, since you aren't trying to modify attributes in associated objects, you are just trying to associate objects to the current object.  One problem with this approach is there is no guarantee the lookup by attributes is unique, and if not you end up with nondeterministic behavior.

You can mostly do what you want with the following code

```ruby
album = Album.create(name: 'The Second Album')
album.add_artist(Artist.find_or_create(name: 'Existing artist or not'))
```

I think if Sequel was to support a feature that makes this easier, it should be through a separate plugin, not through nested_attributes.  It might be best to work on an external plugin that does that, and if it becomes popular open up a thread on sequel-talk about shipping it with Sequel.

jeremyevans:
Looks like my code example was wrong, should be:

```ruby
album = Album.create(name: 'The Second Album',
  :artist=>Artist.find_or_create(name: 'Existing artist or not'))
```

But the previous example would work if album and artist were reversed.

AlexWayfer:
@jeremyevans OK, thanks.
The thing is that I get one big Hash with 5-6 associations, some even nested, and don't want to manually convert each with `find_or_create`.
But thanks, I'm already writing a separate common code, maybe even make the extension (gem) from it.

PR #1167, opened 2016-04-07
Add optional block parameter to dataset hash methods

mwpastore:
Using to_hash (or select_hash, or _groups) to build quick lookups from datasets and models is extremely handy, but I find myself writing constructs like this:

```ruby
lookup = Hash.new do |hash, key|
  hash[key] = MyModel.new :key => key, :value => "unknown"
end.merge! MyModel.to_hash(:key)
```

or:

```ruby
lookup = Hash.new(["unknown"])
  .merge! DB[:my_table].select_hash_groups(:key, :value)
```

This patch adds an optional `:hash` parameter to the various hash methods, into which entries are merged. This allows slightly more readable versions of the above:

```ruby
lookup = MyModel.to_hash :key, nil, :hash => Hash.new do |hash, key|
  hash[key] = MyModel.new :key => key, :value => "unknown"
end
```

and:

```ruby
lookup = DB[:my_table].select_hash_groups(:key, :value, :hash => Hash.new(["unknown"]))
```

I did not carry the change through to the StaticCache plugin. 

Methods affected:
* Dataset#to_hash
* Dataset#to_hash_groups
* Dataset#select_hash
* Dataset#select_hash_groups
* Model#to_hash

### DIFF
--- a/lib/sequel/dataset/actions.rb
+++ b/lib/sequel/dataset/actions.rb
@@ -558,7 +558,9 @@ module Sequel
     end
     
     # Returns a hash with key_column values as keys and value_column values as
-    # values.  Similar to to_hash, but only selects the columns given.
+    # values.  Similar to to_hash, but only selects the columns given.  Like
+    # to_hash, it accepts an optional :hash parameter, into which entries will
+    # be merged. 
     #
     #   DB[:table].select_hash(:id, :name) # SELECT id, name FROM table
     #   # => {1=>'a', 2=>'b', ...}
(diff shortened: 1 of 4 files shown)


jeremyevans:
Thanks for the patch!  This feature looks good, but I suggest one change to the API.  Instead of passing a block that gets passed to `Hash.new`, add an options hash to the methods, and accept a `:hash` option that specifies the hash to use.  That way you can pass in a hash with a default value, a hash with a default proc, or any object that supports `#[]` and `#[]=`.  This seems more flexible than just using a block.  One advantage over the current approach is that hashes with default procs are not marshallable, so forcing a default proc when the default value is static (as in `["unknown"]` above) limits your ability to marshal the hash.

mwpastore:
Hrrm, I prefer the idiomatic block style, but I see your points. Would you be opposed to allowing both?

jeremyevans:
I don't think it makes sense to support both.  One reason not to is that it allows us to use the block differently in the future, such as for transforming the input.  If we start using the block as the Hash.new block, we are forever restricted on using it for something else.  Considering how rarely this is needed, that doesn't seem like a good tradeoff.

mwpastore:
Alrighty, I've made the change and updated the PR description. Thanks @jeremyevans! Let me know if you'd like to see any additional changes or tests.

jeremyevans:
Thanks, these changes look good.  I'll merge and test tomorrow.

PR #1169, opened 2016-04-09
Correct reference to create_or_update in UpdateOrCreate docs

petedmarsh:
This is an extremely minor change, happend to notice that there was a reference to `create_or_update` where it should have been `update_or_create` in the documentation for the UpdateOrCreate plugin.


jeremyevans:

PR #1170, opened 2016-04-09
Add validates_operator validator

petedmarsh:
This checks that an attribute compares favourably against a specified
value for a given operation, e.g.

    class Example < Sequel::Model
      def validate
        super
        validates_operator(:>, 3, :something)
      end
    end

Feature discussion:

https://groups.google.com/forum/#!topic/sequel-talk/EMrbjLFrJpA


jeremyevans:
Thanks for the patch! This looks good.  I'll try to test and merge on Monday.

PR #1174, opened 2016-04-22
Register custom range types per Database

steveh:
Original discussion: https://groups.google.com/forum/#!topic/sequel-talk/kCZD4nMAdpY

Based on the work in pg_array.

Allows custom range types to be defined on a particular database instead of globally.

Supports looking up the oid and subtype oid from the pg_range table.

Also corrected an error message where a converter and subtype_oid were both provided.


jeremyevans:
This looks good and is exactly what I had in mind.  I'll try to merge and test in a few days (currently traveling for conferences), but this should definitely make the next release.  Thanks!

PR #1177, opened 2016-05-03
Add option -T to filter selected tables when using -C option

pedrobranco:
Added **new option -T** which is used along with -C option, allowing to **copy only the selected tables to the new database** (all tables are separated with a comma, if you need to specify any WHERE clause you can put a pipe '|' with following a condition).

Example:
```bash
./sequel/bin/sequel -C -T "table1,store|storeid='123'" postgres://postgres:postgres@localhost/exampledb sqlite://./example-output.db
```

### DIFF
--- a/bin/sequel
+++ b/bin/sequel
@@ -18,6 +18,26 @@ load_dirs = []
 exclusive_options = []
 loggers = []
 
+# To use in option -T
+condition_separator = '|'
+only_selected_tables = false
+selected_tables = Array.new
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch.  Unfortunately, I don't think this is a good fit for bin/sequel.  It's a slippery slope (what if I only want specific columns for certain tables?).  If you want to customize how the database copying is done, you should probably write your own program that uses Sequel.

PR #1183, opened 2016-05-08
Change Sequel::Model#lock! to accept a lock style

petedmarsh:
Previously lock! locked the row the model represented with whatever
Dataset#for_update resolved to for the underlying database. This
changes it to accept an optional style of lock that will be aquired
instead (working as Dataset#lock_style does):

    e = Example.first
    e.lock!
    #=> SELECT * FROM examples WHERE id = 1 FOR UPDATE LIMIT 1

    e = Example.first
    e.lock!('FOR NO KEY UPDATE')
    #=> SELECT * FROM examples WHERE id = 1 FOR NO KEY UPDATE LIMIT 1

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,8 @@
 
 * Allow json_serializer :include option with cascaded values to work correctly when used with association_proxies (jeremyevans)
 
+* Allow Sequel::Model#lock! to accept an (optional) lock style (similarly to Dataset#lock_style) (petedmarsh)
+
 === 4.34.0 (2016-05-01)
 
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch.  This looks good, but you can simplify it by having the argument default to `:update`, and just calling `lock_style` with the argument.  I'll merge and test tomorrow.

petedmarsh:
I'll refactor this later this evening :)

petedmarsh:
@jeremyevans refactored, thank you for the review!

jeremyevans:
Looks good, I'll try merging and testing shortly.

PR #1188, opened 2016-05-20
Add Postgres Bignum -> bigserial fix

docelic:
Do convert Bignum to bigserial on Postgres, but
only if the default value has not been specified.

Otherwise, Postgres errors out with:

ERROR:  multiple default values specified for column "..." of table "..."


jeremyevans:
Can you give an explanation of why you want this, show the code you are using, or include a spec?  It sounds like you are doing something like `primary_key :id, :type=>Bignum, :default=>value`, which I wouldn't expect to work (switch to `Bignum :id, :primary_key=>true, :default=>value` in that case),

docelic:
Indeed. Great, thanks!

PR #1195, opened 2016-05-24
add YARD configuration to improve documentation compability

skrobul:
While this project natively uses RDoc documentation format, it is possible to
use it with [YARD][1] without any major modifications.

One of the reasons why it might be useful is usage with offline documentation
tools that use YARD to generate HTML documentation from both RDoc and YARD input.
One popular example of such applications is [Dash][2].

Prior to this commit, the documentation converted that way included only
information that was embedded directly in the code, and all useful documents
from `doc/*.rdoc` were missing.

[1]: http://yardoc.org/
[2]: https://kapeli.com/dash

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@
 *.swp
 /coverage
 /rdoc
+/doc/**/*.html
+/doc/css
+/doc/js
+/.yardoc
 /sequel-*.gem
 /spec/bin-sequel-*
 /spec/spec_config.rb
--- a/.yardopts
+++ b/.yardopts
@@ -1,0 +1,6 @@
+--no-private
+--title "Sequel: The Database Toolkit for Ruby"
+lib/**/*.rb
+-q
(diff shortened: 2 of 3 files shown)


jeremyevans:
Thanks for the patch.  Unfortunately, but I don't think this is a good idea. Sequel doesn't use YARD and I do not want to add YARD-specific code to it.  I think such a path leads to a slippery slope.

skrobul:
@jeremyevans Thanks for prompt response. I totally understand and respect your decision. I attempted to make it work with just RDoc (`.document`) but wasn't successful. It's a shame, because sites that are showing up as first results in google for 'sequel documentation' (i.e. http://rdoc.info) don't have most of the docs.

jeremyevans:
The first Google result for "sequel documentation" for me is http://sequel.jeremyevans.net/documentation.html, which should have all of the docs.

PR #1198, opened 2016-05-27
Transaction refactoring

chanks:
This is the transaction refactoring required to support #1197 as an extension.

The extension itself is at https://github.com/chanks/sequel-savepoint-hooks

### DIFF
--- a/lib/sequel/database/misc.rb
+++ b/lib/sequel/database/misc.rb
@@ -170,7 +170,7 @@ module Sequel
       synchronize(opts[:server]) do |conn|
         if h = _trans(conn)
           raise Error, "cannot call after_commit in a prepared transaction" if h[:prepare]
-          add_hook(conn, :after_commit, block)
+          add_transaction_hook(conn, :after_commit, block)
         else
           yield
         end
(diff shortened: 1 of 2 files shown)


lib/sequel/database/misc.rb
```diff
@@ -341,6 +341,13 @@ def url
     def adapter_initialize
     end
 
+    # Set the given callable as a hook to be called. Type should be either
+    # :after_commit or :after_rollback.
+    def add_hook(conn, type, block)
```
jeremyevans: We should probably rename this `add_transaction_hook`

lib/sequel/database/transactions.rb
```diff
@@ -198,6 +198,8 @@ def _trans(conn)
     def add_transaction(conn, opts)
       hash = {}
 
+      apply_transaction_options(conn, opts, hash)
```
jeremyevans: I think it would be better to change this method so that it returned the hash to use (possible name: `transaction_state_hash`), instead of relying on side-effects.

lib/sequel/database/transactions.rb
```diff
@@ -299,15 +306,23 @@ def connection_execute_method
       :execute
     end
 
+    # Retrieve the transaction hooks that should be run for the given
+    # connection and commit status.
+    def get_hooks(conn, committed)
```
jeremyevans: This should probably be renamed to `transaction_hooks`

chanks:
Made the changes you brought up so far.

jeremyevans:
Thanks! The changes look good, let me test them out.  Assuming no problems I'll merge later today.

chanks:
Thank you!

PR #1205, opened 2016-06-22
Added a database-specific migration example to the migrations doc.

onewheelskyward:


jeremyevans:
This looks OK, but `database_type` returns a symbol, you probably want to use `== :mysql` instead of `=== 'mysql'`.

jeremyevans:
I went ahead and merged this, then added a commit to change the example and some of the text.

PR #1208, opened 2016-06-27
Add connection_expiration extension for automatically disconnecting c…

pdrakeweb:
…onnections after an expiration time.

This is useful in cases where the recommendation is not to maintain open connections for an extended period of time such as with RDS (see https://forums.aws.amazon.com/message.jspa?messageID=546388#546388).

### DIFF
--- a/lib/sequel/extensions/connection_expiration.rb
+++ b/lib/sequel/extensions/connection_expiration.rb
@@ -1,0 +1,88 @@
+# frozen-string-literal: true
+#
+# The connection_expiration extension modifies a database's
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch. It seems odd to add an extension just to work around problems in one provider's IaaS implementation.  That being said, the implementation looks good (it appears to be mostly copied from the connection_validator extension, but that's fine) and the specs look complete.  I suppose this could be useful in other cases outside of AWS, so I'll test this tomorrow and assuming no problems I'll merge it.

PR #1209, opened 2016-06-29
Fix disconnect bug in connection expiration extension and add unit test.

pdrakeweb:


jeremyevans:
Looks good, sorry I missed it in my review.

PR #1215, opened 2016-07-12
Make MigrationAlterTableReverser work with add_constraint with name in a hash

soupmatt:
If you are using postgresql and you want to have a reversible migration with an `add_constraint` and `:not_valid => true` like this:

```ruby
Sequel.migration do
  change do
    alter_table :my_table do
      add_constraint({:name => :new_check, :not_valid => true}, 'col_a > 5')
    end
  end
end
```

You will get an error like this: `PG::UndefinedObject: ERROR:  constraint "{:name => :new_check, :not_valid => true}" of relation "my_table" does not exist`

This patch fixes this.

### DIFF
--- a/lib/sequel/extensions/migration.rb
+++ b/lib/sequel/extensions/migration.rb
@@ -245,7 +245,9 @@ module Sequel
     end
 
     def add_constraint(*args)
-      @actions << [:drop_constraint, args.first]
+      name = args.first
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch!  This looks good, I'll try to merge and test tomorrow.

PR #1216, opened 2016-07-13
Make Dir#each deterministic

viktors:
Dir#each can return entries in different order on different systems. Make the order deterministic by sorting entries first.
An example of the problem: two timestamped migrations with the same timestamp could be applied in different order on development and production.

### DIFF
--- a/lib/sequel/extensions/migration.rb
+++ b/lib/sequel/extensions/migration.rb
@@ -402,7 +402,7 @@ module Sequel
     # after 2005, otherwise uses the IntegerMigrator.
     def self.migrator_class(directory)
       if self.equal?(Migrator)
-        Dir.new(directory).each do |file|
+        Dir.new(directory).sort.each do |file|
           next unless MIGRATION_FILE_PATTERN.match(file)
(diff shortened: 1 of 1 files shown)


jeremyevans:
If two migrations have the same version, I would expect the order to be undefined, and I don't think it's a good idea to define it.  If you have two migrations with the same timestamp, and you are depending on which order the migrations run, you should problem change things.

Note that the timestamp migrator already allows migrations to be applied in a different order in different environments, even if the timestamps are not identical.  That's actually the whole point of the timestamp migrator.  If you have a completely defined order for migrations, the integer migrator is better (and does not suffer from this issue).

Example:

Development has threes branches, master, A, and B.  B adds migration 20160712_foo.rb.  A adds migration 20160713_bar.rb.  Developer merges branch A into master and runs migrations, then merges branch B into master and runs migrations.  Result is 20160712_foo.rb is run after 20160713_bar.rb.  Developer than pushes master into production and runs migrations.  Result is 20160712_foo.rb is run before 20160713_bar.rb.

PR #1219, opened 2016-07-16
Load SQLite extension functions

karfas:
A small change allowing to dynamically load SQLite extension functions.

### DIFF
--- a/lib/sequel/adapters/sqlite.rb
+++ b/lib/sequel/adapters/sqlite.rb
@@ -114,7 +114,10 @@ module Sequel
         if opts.has_key?(:extensions)
           ext = opts[:extensions]
           db.enable_load_extension(1)
-          ext.each { |file| db.load_extensison file }
+          ext.each do |file|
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch. We don't usually accept new features without automated tests, but I guess testing this would require shipping an SQLite extension, which I don't want to do.  Could you please put together a small example of using this that I could use to do some manual testing?

jeremyevans:
I should note that you can already do this fairly easily using an after_connect proc:

```ruby
DB = Sequel.sqlite(:after_connect => proc do |conn|
  conn.enable_load_extension(1)
  conn.load_extension(file)
  conn.enable_load_extension(0)
end)
```

Considering that, and the fact that no example has been posted here, I'm going to close this.  If you still think this is something Sequel should support, please post on the sequel-talk Google Group and see if there is community support for your proposal.

karfas:
Thank you very much for the information.
You are right, this works also in the :after_connect proc. 

We can forget the patch.

Thanks again,
Thomas

PR #1221, opened 2016-07-25
Regular expression support for Oracle 10g+ via REGEXP_LIKE

johnhcaldwell:
Pieced together using modified functions/snippets from the mysql and postgres adapters.  Smart about which versions of Oracle are supported (10g or later.)

### DIFF
--- a/lib/sequel/adapters/shared/oracle.rb
+++ b/lib/sequel/adapters/shared/oracle.rb
@@ -90,6 +90,23 @@ module Sequel
           count > 0
       end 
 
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch! This looks good.  The only change I would make is that `sql << ", #{literal('i')}"` should be either `sql << ", 'i'"` or `sql << ", "; literal_append(sql, 'i')`.

I'll try testing this tomorrow, assuming no problems it should be merged and will make the next release.

PR #1224, opened 2016-08-03
Removed `.first` for `call_mssql_sproc` to allow multiple results

MvanDiemen:


jeremyevans:
The `first` call returns the results, numrows, and potentially other variables if using output parameters.  This API is not designed to return more than a single row.  In any case, removing the `first` call changes the API in a backwards incompatible manner, and as such cannot be accepted.  

Additionally, as stated in the contribution guidelines, please do not mix whitespace changes with code changes.

MvanDiemen:
Yeah, I created the PR a bit to fast, I will continue on this a bit to make it both backwards compatible and according to the contribution guidelines.

However, the `first` call only returns the first item of the result, even when there are multiple result items.

If you mean the stored procedure with "API", than I am afraid I have to disagree. If you are talking about Sequel, I can understand but the stored procedure is not limited to just 1 result item.

MvanDiemen:
https://gist.github.com/MvanDiemen/1fae5a3bf3cdfb2760eaa8d6c4fe35bf

jeremyevans:
I am referring to Sequel's API.  We can't change from returning a hash to returning a dataset.  It is possible to add a new method that returns a dataset (e.g. `mssql_sproc_ds`), and have the existing method call that, though.  If you'd like to do that, please make sure to include specs for such a method with a new pull request.

PR #1225, opened 2016-08-03
Added `#mssql_sproc_ds` method

MvanDiemen:
To be able to use multiple results from a Stored Procedure this method is added that returns the entire dataset.

### DIFF
--- a/lib/sequel/adapters/shared/mssql.rb
+++ b/lib/sequel/adapters/shared/mssql.rb
@@ -141,7 +141,7 @@ module Sequel
 
         ds = dataset.with_sql(sql)
         ds = ds.server(opts[:server]) if opts[:server]
-        ds
+        ds.all
       end
 
       # Microsoft SQL Server uses the :mssql type.
--- a/lib/sequel/adapters/shared/mssql.rb
+++ b/lib/sequel/adapters/shared/mssql.rb
@@ -38,6 +38,8 @@ module Sequel
 
       # Execute the given stored procedure with the given name.
       #
+      # This method returns a single hash. See call_mssql_sproc.
+      #
       # Options:
       # :args :: Arguments to stored procedure.  For named arguments, this should be a
       #          hash keyed by argument named.  For unnamed arguments, this should be an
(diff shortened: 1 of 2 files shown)


jeremyevans:
Instead of returning a dataset, this returns an array, so a name like `mssql_sproc_ds` doesn't make sense.  I think instead of using `all` and returning an array, you should just return the dataset itself.  That's more flexible as it allows the user to use dataset methods.

What happens when the stored procedure returns multiple rows that have different columns?

MvanDiemen:
Reverted to the dataset itself.

A stored procedure should not be able to return multiple rows with different columns, or the differentiating columns should be nil for the other row.

jeremyevans:
In general, Sequel doesn't support running a single command returning multiple result sets, which is what this appears to allow.  The specs you added don't appear to actually use this feature with a stored procedure that returns results.  Can you include a more appropriate spec that shows how this is designed to be used (returning multiple rows), and actually checks the result values instead of just the type?

MvanDiemen:
I understand the decision to only support single commands with single results, but I don't know any other options to get a second result from a stored procedure as it's not something you can get any other way. I'll update the specs a bit.

jeremyevans:
I think the issue here is that Sequel doesn't (and won't) support a single query returning multiple result sets (multiple rows is fine).  This isn't specific to Microsoft SQL Server, this is how Sequel in general works, as the API is based around datasets yielding hashes where the keys are the same for all of the dataset's hashes.  Any time that multiple result sets are used, you cannot ensure that.

Note that you can work around this issue by dropping down to the driver level using `Database#synchronize` and running the appropriate code directly against the underlying driver (assuming the underlying driver supports it).

MvanDiemen:
I am not trying to add support for multiple result sets, but I am trying to get multiple rows with the same keys.

When running my stored procedure, `numrows`, `result` and `output` are all nil, whilst the call itself only returns the first of the two rows found.

MvanDiemen:
However when I run the command with `DB.fetch("execute my_sproc").all It returns the dataset itself.

jeremyevans:
My guess is that's because even if you are not using multiple result sets inside the stored procedure, Sequel is running a query to select a separate result set after the stored procedure.  So if your stored procedure returns any results, you end up with multiple result sets.

Basically, `call_mssql_sproc` does not work with stored procedures that return results.  If you are just looking to return the results of a stored procedure, that `fetch.all` command should work (it should return an array, not a dataset).

MvanDiemen:
Ah yeah, ofcourse, that's an Array.

It makes sense to keep the `call_mssql_sproc` to only return one item, so I think the `fetch.all` is the most sane (on both the API as the client side) option.

Thanks for the help and sparring!

Issue #1184, opened 2016-05-11
Multiple warnings


fornellas:
### Complete Description of Issue

Sequel gives a lot of warnings while running. I usually execute my tests with warnings enabled, and Sequel is giving a lot of unneeded output. They seem harmless, but surely are noisy.

### Simplest Possible Self-Contained Example Showing the Bug

```
$ irb -rsequel -w
/home/fornellas/local/gems/23/gems/wirble-0.1.3/lib/wirble.rb:410: warning: assigned but unused variable - nocol
/home/fornellas/local/gems/23/gems/wirble-0.1.3/lib/wirble.rb:440: warning: assigned but unused variable - colors
/home/fornellas/local/gems/23/gems/wirble-0.1.3/lib/wirble.rb:78: warning: instance variable @verbose not initialized
/home/fornellas/local/gems/23/gems/awesome_print-1.6.1/lib/awesome_print/core_ext/array.rb:19: warning: method redefined; discarding old -
/home/fornellas/local/gems/23/gems/awesome_print-1.6.1/lib/awesome_print/core_ext/array.rb:19: warning: method redefined; discarding old &
>> Sequel.extension :migration
=> [:migration]
>> Sequel.sqlite('/tmp/a')
=> #<Sequel::SQLite::Database: {:adapter=>:sqlite, :database=>"/tmp/a"}>
>> db = Sequel.sqlite('/tmp/a')
=> #<Sequel::SQLite::Database: {:adapter=>:sqlite, :database=>"/tmp/a"}>
>> Sequel::Migrator.run(db, '/tmp/m')
/home/fornellas/local/gems/23/gems/sequel-4.34.0/lib/sequel/dataset/query.rb:80: warning: instance variable @columns not initialized
/home/fornellas/local/gems/23/gems/sequel-4.34.0/lib/sequel/dataset/query.rb:80: warning: instance variable @columns not initialized
/home/fornellas/local/gems/23/gems/sequel-4.34.0/lib/sequel/dataset/query.rb:80: warning: instance variable @columns not initialized
/home/fornellas/local/gems/23/gems/sequel-4.34.0/lib/sequel/dataset/misc.rb:85: warning: instance variable @frozen not initialized
/home/fornellas/local/gems/23/gems/sequel-4.34.0/lib/sequel/dataset/query.rb:80: warning: instance variable @columns not initialized
[...]

PR #1227, opened 2016-08-25
Remove 'not initialized instance variable' warning

davydovanton:
Hello,
I use `Sequel::Model` in simple sinatra app and when I call tests I see warning about instance variable like this:
```
User
  validation
    when user does not have name
../gems/sequel-4.37.0/lib/sequel/model/base.rb:2307: warning: instance variable @singleton_setter_added not initialized
      should not be valid
```

That's why I created this PR which fixed this warning.

_P.S.: should I create some tests for this?_

### DIFF
--- a/lib/sequel/model/base.rb
+++ b/lib/sequel/model/base.rb
@@ -1290,6 +1290,7 @@ module Sequel
       #     a.name = 'Bob'
       #   end
       def initialize(values = {})
+        @singleton_setter_added = false
         @values = {}
         @new = true
         @modified = true


jeremyevans:
By design, Sequel does not initialize instance variables, as doing so reduces performance.  See https://github.com/jeremyevans/sequel/issues/1184, https://groups.google.com/forum/#!searchin/sequel-talk/warning/sequel-talk/9aehjtVM6II/6BpInBub1LcJ and https://groups.google.com/forum/#!msg/sequel-talk/Zc-PreRU-o8/xuF2rMd7BwAJ

davydovanton:
got it, thanks 👍

PR #1228, opened 2016-08-30
Add #paged_each to odbc dataset for loading large datasets

myers:
When loading large datasets, we don't want to load all the rows in at once, but rather load them in a chunk at a time.  This follows the same API as in Postgres::DataSet's #paged_each.

### DIFF
--- a/lib/sequel/adapters/odbc.rb
+++ b/lib/sequel/adapters/odbc.rb
@@ -93,13 +93,17 @@ module Sequel
 
       Database::DatasetClass = self
 
+      def paged_each(opts=OPTS, &block)
+        clone({:rows_per_fetch=>1000}.merge!(opts)).each(&block)
+      end
+
       def fetch_rows(sql)
         execute(sql) do |s|
           i = -1
           cols = s.columns(true).map{|c| [output_identifier(c.name), c.type, i+=1]}
           columns = cols.map{|c| c.at(0)}
           self.columns = columns
-          if rows = s.fetch_all
+          fetch_raw_rows(s) do |rows|
             rows.each do |row|
               hash = {}
               cols.each{|n,t,j| hash[n] = convert_odbc_value(row[j], t)}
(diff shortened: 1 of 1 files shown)


jeremyevans:
First, thanks for the patch!

In the postgres adapter, a cursor is used, so it is a little different than how this is handled. That is, unless ODBC is specified to always use a cursor internally in this case, and I'm not sure of that. If you can find ODBC documentation that specifies that, please let me know.  Using `fetch_many` like this may only handle the last level of conversion, it still may be retrieving all rows for the query when the first `fetch_many` is called, and just buffering them, depending on the ODBC driver.

Can you let me know which ODBC drivers you have tested this with, and how you confirmed that this only loads records from the database using the rows_per_fetch specified?  I'm hesitant to make this the default behavior of the odbc adapter (unless ODBC is specified to always use a cursor internally for it), but it's possible it could be the default behavior in an odbc subadapter.

myers:
@jeremyevans Thanks for your quick reply.  I'm currently using this with the ODBC driver for http://www.snowflake.net/ .  I have confirmed that it's only loading a subset because with this PR applied linux is no longer OOM-killing my processes :).  I agree it's hard to know what's going on under the ODBC API, but it's working in this case.  

I don't have access to many databases with ODBC drivers.

I'm not attempting to make this the default behavior.   Here's an example:

```
MyDB["show tables"].each {|r| pp r}
```

With or without the PR is would load all the rows at once.

```
MyDB["show tables"].paged_each {|r| pp r}
```

This method is added by the PR and will load the rows 1000 at a time.

Perhaps a solution is to document this method as experimental and ask for feedback on if it works correctly on different databases?

jeremyevans:
Unfortunately, `paged_each` is part of Sequel's Dataset API and has been for a long time, I can't just mark is as experimental on a single adapter. 

Note that `paged_each` is not added by your pull request, `paged_each` is defined on Sequel::Dataset, what this pull request does is override it and give it different behavior.  Have you tried `paged_each` without this patch?

Based on what you've said, I don't think this could be merged, it is too likely to change existing behavior for other users of the odbc adapter.  However, I'm open to having `fetch_rows` call `fetch_raw_rows` , and then you could easily add the behavior you want using a dataset extension that overrides `paged_each` and `fetch_raw_rows`.  Do you think that would be acceptable?

jeremyevans:
Closing due to lack of response.  I'm still open to adding `fetch_raw_rows` to the odbc adapter and having `fetch_rows` call it, please submit a pull request if you'd like that to be added.

Issue #1232, opened 2016-09-15
The pg_interval extension doesn't support literaliizing the hour or week components of ActiveSupport::Duration objects.


chanks:
ActiveSupport::Duration supports 7 date parts: `[:years, :months, :weeks, :days, :hours, :minutes, :seconds]`
Sequel's pg_interval extension only recognizes 5 of them: `[:years, :months, :days, :minutes, :seconds]`

This results in incorrect behavior when literalizing ActiveSupport::Duration objects that use the `hour` or `week` parts:

```ruby
[1] > Sequel::Postgres::IntervalDatabaseMethods.literal_duration(2.months)
=> "2 months "
[2] > Sequel::Postgres::IntervalDatabaseMethods.literal_duration(2.years)
=> "2 years "
[3] > Sequel::Postgres::IntervalDatabaseMethods.literal_duration(2.hours)
=> "0"
[4] > Sequel::Postgres::IntervalDatabaseMethods.literal_duration(2.weeks)
=> "0"
[5] > Sequel::Postgres::IntervalDatabaseMethods.literal_duration(2.seconds)
=> "2 seconds "
```

I'll have some time tomorrow to work up a PR for this, if you like?

Thanks!

jeremyevans:
This is due to a change in behavior in ActiveSupport 5.  Both ActiveSupport 3 and ActiveSupport 4:

```
irb(main):001:0> 2.weeks
=> 14 days
irb(main):002:0> 2.hours
=> 7200 seconds
```

Now, this is mentioned prominently in the ActiveSupport 5 release notes, so I'm not saying this is ActiveSupport's fault, but it certainly wasn't always broken.  Anyway, please do submit a PR for this if you have time.  If you don't get around to it, I should be able to take care of it before the next release.

chanks:
Yeah, definitely didn't mean to imply it was always broken, I'd just never tried out the extension until now :smile:.

I opened PR #1233, so closing this.

PR #1233, opened 2016-09-16
Support weeks and hours in pg_interval extension.

chanks:
See #1232. This fix is pretty simple and only handles literalization, to remain backwards compatible with earlier versions of ActiveSupport. I suppose it'd be nice to tweak the parsing to populate the week and hour parts, but it'd require checking the current version of ActiveSupport supports that, and felt like it wouldn't really be worth the trouble.


jeremyevans:
These changes look good, I'll merge and test shortly.

PR #1234, opened 2016-09-16
Supports AFTER and FIRST MySQL column positioning options for migrations

AnthonyBobsin:
This PR allows columns to be added at a specific position within a table row by using `AFTER` and `FIRST` when using `alter_table(:some_table) { add_column ... }` MySQL schema generators.

### DIFF
--- a/doc/schema_modification.rdoc
+++ b/doc/schema_modification.rdoc
@@ -339,6 +339,11 @@ hash:
     add_column :copies_sold, Integer, :default=>0
   end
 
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch!  I'll probably make a few tweaks, but this is a great start.

AnthonyBobsin:
No problem. I love what you have built here, just happy I can contribute!

That sounds great, please let me know if I can help at all with the tweaks as well.

PR #1236, opened 2016-09-22
Fix minor typo in transaction docs

theoretick:


jeremyevans:

PR #1237, opened 2016-09-25
fix grammatical error in documentation

darrenburgess:


jeremyevans:

PR #1238, opened 2016-09-29
Add block support to Sequel::Model::Errors#add

Wardrop:
This is something I've been monkey patching in my projects, and figured it's probably worth adding to Sequel.

Here's the idea. Often when validating fields, a single field may have many validations. For example you may first check the format is correct, and then check that the value is within range; there's no use checking if the value is within range if it's not even in the right format to begin with. The block provides a convenient mechanism for this common situation. Refer to the following contrived example:

``` ruby
errors.add(:start_date, "Start date must be a valid date") { DateTime === start_date }
errors.add(:start_date, "Start date must not be in the past") { start_date >= DateTime.today }
```

In the example above, the second validation is only checked is there's not already an error associated with the `start_date` field. The block defines the valid condition, so it it returns false, the validation has failed, and the error message is appended.

Let me know if you would like me to add unit tests for this change.

### DIFF
--- a/lib/sequel/model/errors.rb
+++ b/lib/sequel/model/errors.rb
@@ -7,11 +7,16 @@ module Sequel
     class Errors < ::Hash
       ATTRIBUTE_JOINER = ' and '.freeze
 
-      # Adds an error for the given attribute.
-      #
+      # Adds an error for the given attribute. If a block is given, only adds the error if the block returns false,
+      # and there are no other errors already associated with the attribute.
       #   errors.add(:name, 'is not valid') if name == 'invalid'
       def add(att, msg)
-        fetch(att){self[att] = []} << msg
+        att_errors = fetch(att){self[att] = []}
+        if block_given?
+          att_errors << msg if self[att].empty? && !yield
+        else
+          att_errors << msg
+        end
       end
 
       # Return the total number of error messages.


jeremyevans:
Thank you for submitting this pull request.  Unfortunately, I do not think that hiding a conditional in a block as this does is a good design.  Your example can be reduced to:

```ruby
if start_date.is_a?(DateTime)
  if start_date >= DateTime.today
    errors.add(:start_date, "Start date must not be in the past")
  end
else
  errors.add(:start_date, "Start date must be a valid date")
end
```

which I think is more clear than the block form, even if it is a bit more verbose.  I don't find the block form intuitive, and I think not adding an error for the column if the error is already present is also a bad default.  There is a reason that `errors.on(:column)` returns an array.

Note that it is fairly easy to support the API you want using an external plugin, and I would recommend creating one if you want to use such an API.

Wardrop:
Fair enough, I'll look into implementing it as a plugin.

In some of the applications I've had, there could be up to 4 or 5 validations per field, each subsequent validation assuming the validation before succeeded. Say you've got 10 fields averaging 3 validations each (worst case I guess), that would be quite a mess of nested if/else statements. With that said, perhaps as you said, adding a block to the `add` method may not be the most intuitive. Perhaps a new method named something like `add_unless` would be more intuitive. I'll implement this as a plugin for myself.

Wardrop:
Here's what I've come up with, if anyone is interested:

``` ruby
module Sequel
  module Plugins
    module ExtendedErrors
      class Errors < Sequel::Model::Errors
        # Evaluates the given block before adding error. Only executes the block (and adds the method) if the field doesn't currently have any errors.
        # The given block must return _falsey_ for the error to be added.
        def add_unless(att, msg)
          raise "Block must be provided" unless block_given?
          add(att, msg) if (!self[att] || self[att].empty?) && !yield
        end
      end

      module InstanceMethods
        def errors_class
          Sequel::Plugins::ExtendedErrors::Errors
        end
      end
    end
  end
end
```

PR #1239, opened 2016-09-29
Fix minor typo in object model docs.

conradbeach:


jeremyevans:
Awesome, thanks for this!

PR #1240, opened 2016-10-03
Support specifying the predicate of a partial unique index in Dataset#insert_conflict

chanks:
I have a partial unique index, declared like `index [:a, :b], unique: true, where: {c: true}`. Postgres supports using it in an ON CONFLICT clause by specifying the index predicate as part of the target, e.g. `INSERT ... ON CONFLICT ("a", "b") WHERE ("c" IS TRUE) DO UPDATE ...`. Unfortunately, Sequel doesn't support that part of ON CONFLICT yet, so here's a small PR implementing it.

(Side note: I initially tried to work around this issue by passing a constraint name to `insert_conflict`'s constraint naming option, but unfortunately Postgres doesn't support filters on unique constraints, only unique indexes, and it won't accept being passed an index name rather than a constraint name. So it looks like the only way to support this use case in Sequel is via a patch like this.)

The only downside I can see with this patch is that the `:where` and `:update_where` options might be easily confused by users. I went with `:where` as the option name because it matches the SQL output (and I'm glad that `:update_where` was named as specifically as it was!), and I tried to make its usage clear in the docs. I think renaming the option to something like `:index_where` might be preferable, but erred on the side of matching the generated SQL.

Postgres' INSERT docs, for convenience: https://www.postgresql.org/docs/9.5/static/sql-insert.html

### DIFF
--- a/lib/sequel/adapters/shared/postgres.rb
+++ b/lib/sequel/adapters/shared/postgres.rb
@@ -1407,7 +1407,7 @@ module Sequel
       # ON CONFLICT. With no options, uses ON CONFLICT DO NOTHING.  Options:
       # :constraint :: An explicit constraint name, has precendence over :target.
       # :target :: The column name or expression to handle uniqueness violations on.
-      # :where :: The index filter, when using a partial index to determine uniqueness.
+      # :conflict_where :: The index filter, when using a partial index to determine uniqueness.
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for submitting this.  I agree that we should offer this feature.  However, I think :where as an option name might get confusing since we have :update_where.  I think :conflict_where might be a better option name.   With that change, I think this can go in.

chanks:
Certainly, updated.

PR #1249, opened 2016-10-13
Update and rename README.rdoc to README.md

glaucocustodio:
Add syntax highlighting (:


jeremyevans:
Sorry, but I want to keep the README in rdoc format.  If you have any content changes you would like to make to the README, please submit them as a separate pull request.

glaucocustodio:
Ok, but I think a good project must have a nice documentation and syntax highlighting helps a lot in the readability.

jeremyevans:
The README is syntax hightlighed: http://sequel.jeremyevans.net/rdoc/files/README_rdoc.html

GitHub may not respect the syntax highlighting, but that's an issue with GitHub itself.

glaucocustodio:
Ok, but many people are used to read the README direct on GitHub. 

The ideal would be having syntax highlighted docs in any source (in your site, GitHub etc).

jeremyevans:
Well, contact support@github.com and ask them to enable syntax highlighing for rdoc files.  That way you benefit all projects using rdoc files.

janko-m:
I hope it's ok that I chime in, as someone who has adopted Jeremy's way of generating documentation in one of my own projects.

@glaucocustodio First of all, you'll notice that the documents linked on the website don't point to the equivalent documents on GitHub, but instead point to RDoc-generated versions on the time of the release. This ensures that they don't contain any unreleased changes.

You generate the HTML-versions of these documents using the same way that you're generating them for method comments in Ruby files. You can choose either `rdoc` syntax, or `markdown` syntax. I chose `markdown` syntax, because I'm more comfortable with it, and this allows me to also use GitHub-flavoured fenced code blocks, which don't make any difference for RDoc itself, but makes it so GitHub adds syntax highlighting. The `rdoc` syntax doesn't support fenced code blocks.

I *think* that this is all-or-nothing, either you can use `rdoc` syntax everywhere, or `markdown` syntax everywhere. In that case the only way that Jeremy could have GitHub correctly syntax highlight them is to completely switch everything from `rdoc` to `markdown` syntax, which would obviously be an unimaginable amount of work. Furthermore, `markdown` syntax has some downsides: it doesn't have the same syntax for generating tables as GitHub (it's arguably [more awkward](https://github.com/janko-m/shrine/blob/f93cbfe7ed24e4ccb02256f216bc29052f302673/lib/shrine/plugins/determine_mime_type.rb#L15-L38)), code examples need 4-spaces ident (while `rdoc` only 2), and I think it also doesn't support explicit links to other classes. Lastly, Jeremy probably prefers the `rdoc` syntax.

Just my explanation from personal experience using the same tools as Jeremy. Since many web apps support automatic detection of language and syntax highlighting it (StackOverflow, Disqus, ...), I think it makes a lot of sense to request the same from GitHub.

PR #1252, opened 2016-10-17
Create a test to reveal issue with quoting in Mock

aguynamedryan:
I think I've found a bug where, even if quote_identifiers_default
returns true for a shared adapter, the mock adapter doesn't quote.

It looks like Mock::Database#quote_identifiers_default is called before
Mock::Database#adapter_initialize so Mock::Database#shared_adapter?
returns nil instead of true

I'm not sure how to fix the bug, but I hope this test proves useful.

### DIFF
--- a/spec/core/mock_adapter_spec.rb
+++ b/spec/core/mock_adapter_spec.rb
@@ -20,6 +20,10 @@ describe "Sequel Mock Adapter" do
           def foo
             @foo
           end
+
+          def quote_identifiers_default
+            true
+          end
         end
 
         module self::DatasetMethods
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the report.  While I agree that this looks like a bug, it's actually not a code bug.  Shared adapters should have their `DatabaseMethods` module extend `Sequel::Database::ResetIdentifierMangling`.  If your shared adapter doesn't do that, that's probably the cause of the problem.  

The reason for this is the identifier mangling really needs to be reset at exactly the point where the `DatabaseMethods` module extends the Database instance, otherwise there can be problems.  If you look at the shared/sub adapters that ship with Sequel, you'll notice that all of them do this.  I apologize that this is not mentioned in the documentation; I think this is a documentation bug.  We don't currently have an adapter writing guide, but I'll add this to the `set_shared_adapter_scheme` method rdoc, since that's the best documentation we currently have on shared adapter creation.

PR #1255, opened 2016-11-09
Changing the warn to print

LIQRGV:
warn will having it's output on stderr which is giving error status on ansible task if we execute it

### DIFF
--- a/lib/sequel/adapters/shared/postgres.rb
+++ b/lib/sequel/adapters/shared/postgres.rb
@@ -176,9 +176,9 @@ module Sequel
       # having callable values for the conversion proc for that type.
       attr_reader :conversion_procs
 
-      # Add a conversion proc for a named type.  This should be used 
+      # Add a conversion proc for a named type.  This should be used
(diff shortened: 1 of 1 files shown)


jeremyevans:
Then redirect stderr to stdout for the task.  Warnings should be printed to stderr, not stdout.  One way to work around this would be do: `def DB.server_version; 80100 end` directly after creating your Database object.

You may want to consider upgrading your PostgreSQL installation, as versions before 8.2 are ancient and long out of support.  

 Also, in the future, as specified in the contribution guidelines, do not include whitespace changes with code changes.

LIQRGV:
Thanks for the input.
FYI, some database like AWS redshift cant updated to latest version.
Sad.

jeremyevans:
If you are using redshift and not PostgreSQL, you may want to use a redshift-specific adapter such as https://github.com/remind101/sequel-redshift

LIQRGV:
Thanks bro!!!
Thats helpfull....

PR #1259, opened 2016-11-18
Fix a memory leak in ADO adapter's execute method

vais:
`ADODB.Connection#Execute` returns an *open* recordset for a row-returning query (and a closed recordset for a non-row-returning query).

Since it's an open recordset, it will remain in memory until it's either closed, or it's associated connection object is closed (which is never, thanks to connection pooling).

(If I understand correctly, ADO adapter's `execute` method is for row-returning queries only, and for non-row-returning queries, `execute_dui` is used instead. Please correct me if I'm wrong, in which case I may need to add a `rescue nil` after `r.close`)


jeremyevans:
Thanks for the patch. This looks good to me, I'll test later today and assuming no problems it will be merged.

PR #1260, opened 2016-11-18
Add timestamps method when creating table

duduribeiro:
Hi,

This PR will add a behave to create default timestamps columns with a syntax sugar.

Instead of defining both columns in all migrations (:created_at, :updated_at),
```
    create_table :cats do
      primary_key :id
      column :created_at, DateTime, null: false
      column :updated_at, DateTime, null: false
```
Is possible to use `timestamps` method, and it will generate both columns
```
    create_table :cats do
      primary_key :id
      timestamps
```

I don't know if this PR is a valid idea for Sequel 😄 .
I tried to find other PR with this behave, and didn't find any result.

Thanks so much

🍻

### DIFF
--- a/lib/sequel/database/schema_generator.rb
+++ b/lib/sequel/database/schema_generator.rb
@@ -288,6 +288,12 @@ module Sequel
         index(columns, opts.merge(:type => :spatial))
       end
 
+      # Add fields for timestamp (:created_at, :updated_at as DateTime)
+      def timestamps
+        column(:created_at, DateTime, null: false)
+        column(:updated_at, DateTime, null: false)
+      end
+
       # Add a unique constraint on the given columns to the DDL.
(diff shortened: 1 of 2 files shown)


jeremyevans:
Sorry, but I don't want helper methods like this in the base schema generators (you can add them via an external extension if you want).  They encourage you to add such columns to all tables, where I think you should think carefully about whether you actually need both such columns, and only add the columns you really need.

In terms of implementation, created_at and updated_at should always have a :default of Sequel::CURRENT_TIMESTAMP, and you should probably use a trigger to set the updated_at value (for PostgreSQL, see sequel_postgresql_triggers).

duduribeiro:
This makes total sense @jeremyevans.. 
Thanks for the reply 😄

PR #1262, opened 2016-11-22
Travis: use ruby-2.3.3

olleolleolle:
This chore PR updates the Ruby 2.3 to be 2.3.3.


jeremyevans:
Well, 2.3.3 is only a single bug fix that doesn't affect Sequel, but I suppose there's no harm in this.

PR #1263, opened 2016-11-22
Fix typos in schema_modification.rdoc

stomar:


jeremyevans:
Thanks, all these look good.

PR #1264, opened 2016-11-23
add Boolean support for mssql when do migration

andyliuliming:


andyliuliming:
Hi Jeremy,
we are working on the bosh and cloud foundry for sql server. we need this change, so we do not need to copy paste two set of migration codes for mssql and mysql.

Thanks very much,
Andy

jeremyevans:
Database-independent boolean support already exists via TrueClass and FalseClass, please read the documentation: http://sequel.jeremyevans.net/rdoc/files/doc/schema_modification_rdoc.html

PR #1267, opened 2016-11-24
MSSQL::Database#rename_table does not work across schemas

vais:
I noticed that after running integration specs, there is always one table, `MY.FOO`, left over in the test database. The reason it is not getting cleaned up is because the cleanup code in the `after` block that deletes the `foo` table assumes that this table is in the default schema. But this table is not in the default schema because `DB.rename_table Sequel[:MY][:bar], :foo` means "rename `MY.BAR` to `MY.FOO`", **not** "rename `MY.BAR` to `dbo.FOO`". I.e. renaming does not work across schemas, and the cleanup code needs to take this into account.

### DIFF
--- a/spec/adapters/mssql_spec.rb
+++ b/spec/adapters/mssql_spec.rb
@@ -424,20 +424,16 @@ describe "A MSSQL database" do
 end
 
 describe "MSSQL::Database#rename_table" do
-  after do
-    DB.drop_table?(:foo)
-  end
-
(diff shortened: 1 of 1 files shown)


jeremyevans:
This looks good to me.  I probably cannot test until next Tuesday, but it should be merged then.

vais:
No problem, thanks, Jeremy. Happy holidays!

PR #1269, opened 2016-11-29
specify the name of an association in json_serialize plugin

sensadrome:
add :name option to :include hash to specify the name of the association.

tested in 1.8.7, 1.9.3 and 2.3.1
compared using JSON.parse, because 1.8.7 reversed the keys in the hash ;)

added the example in the rdoc comments (but I guess the doc directories are git ignored...). I noticed that the examples don't quite show whats expected (they have json_class as a key) but I followed that style in case I missed something!

### DIFF
--- a/lib/sequel/plugins/json_serializer.rb
+++ b/lib/sequel/plugins/json_serializer.rb
@@ -32,6 +32,13 @@ module Sequel
     #   # => '{"json_class":"Album","id":1,"name":"RF","artist_id":2,
     #   #      "artist":{"json_class":"Artist","name":"YJM"}}'
     #
+    # You can specify a name for a given association by passing <tt>:name</tt>
+    # to the <tt>:include</tt> hash
+    #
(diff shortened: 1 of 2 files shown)


jeremyevans:
This looks good, thanks for the patch.  Unfortunatley, this modifies the hash provided, which is not acceptable.  Maybe we should use a slightly different API for this, such as using an aliased expression: `album.to_json(:include=>{Sequel.as(:artist, :singer)=>{:only=>:name}})` . Basically, if an aliased expression is given, treat the expression as the association name, and the alias as the JSON key.

sensadrome:
oh yes that would be much better, I wasn't liking the arbitrary key in there. Incidentally it does work even if I don't delete the :name param. will try the aliased expression and push that...

sensadrome:
yep that works, and feels *so* much better. Thanks. I love this library... great work

jeremyevans:
Yep, this looks good.  I'm not sure if it will make the next release, but I'll try.

PR #1270, opened 2016-11-30
fix Sequel alias expression when not used in hash context

sensadrome:
the alias expression for an association did not work when used standalone (or in an array), this fixes that - didn't spot it before!

### DIFF
--- a/lib/sequel/plugins/json_serializer.rb
+++ b/lib/sequel/plugins/json_serializer.rb
@@ -330,7 +330,14 @@ module Sequel
                 end
               end
             else
-              Array(inc).each{|c| h[c.to_s] = send(c)}
+              Array(inc).each do |c|
+                key_name = c.to_s
+                if c.is_a?(Sequel::SQL::AliasedExpression)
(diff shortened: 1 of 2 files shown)


jeremyevans:
This looks OK, but please change to using if/else as in d08861d3d6e4553b110e87b777bd644cb73f2d2b, and don't add `json_class` usage back to the specs.

sensadrome:
weird, sorry about that - I thought I had fully merged first. my bad. done as requested...

sensadrome:
and there's the if / else etc.

sensadrome:
If I wanted to go about this whilst using .eager to load the associations could you give me some pointers ? I have a nested structure which is doing some lovely n + 1 queries ;)

jeremyevans:
As long as the association is eager loaded, this should avoid the n+1 issues.  But if you are having issues with that, you should probably post on the Google Group or IRC.  I'll see if I can get this merged quickly.

sensadrome:
yeah or course, sorry -wrong place. thanks man.

PR #1279, opened 2016-12-21
Remove little incomprehension

orangeudav:


jeremyevans:
I'm not sure why the previous code would be hard to understand, but I don't have a problem with this change.

PR #1280, opened 2016-12-22
Remove minor inconsistency when :update_on_create == true

NicWagenaar:
Currently the update time can be a couple of milliseconds later then the create time, which is undesired behaviour when :update_on_create == true. This patch solves that by passing the same time value to both set_create_timestamp and set_update_timestamp.

### DIFF
--- a/lib/sequel/plugins/timestamps.rb
+++ b/lib/sequel/plugins/timestamps.rb
@@ -93,7 +93,8 @@ module Sequel
         def set_create_timestamp(time=nil)
           field = model.create_timestamp_field
           meth = :"#{field}="
-          set_column_value(meth, time||=model.dataset.current_datetime) if respond_to?(field) && respond_to?(meth) && (model.create_timestamp_overwrite? || get_column_value(field).nil?)
+          time = model.dataset.current_datetime if !time
+          set_column_value(meth, time) if respond_to?(field) && respond_to?(meth) && (model.create_timestamp_overwrite? || get_column_value(field).nil?)
           set_update_timestamp(time) if model.set_update_timestamp_on_create?
         end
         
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch! This looks good, I'll merge and test later today.

jeremyevans:
Just FYI, if you use the current_datetime_timestamp extension, with the current code the timestamps are the same, as it will use `CURRENT_TIMESTAMP` in both cases.  This is actually better as it reflects the database time instead of the application time (all timestamps in the same transaction will be the same, at least on PostgreSQL).

jeremyevans:
Looking at the existing code more closely, it should already set the update timestamp to the same time as the create timestamp if it sets the create timestamp, since `time ||=model.dataset.current_datetime` is already present. By moving the `time ||=model.dataset.current_datetime`  outside of the if statement, it looks like this patch just attempts to handle the corner case where `set_create_timestamp` is called without arguments, `create_timestamp_overwrite?` is false (the default), and there is already a create timestamp set.  In that case it will set `time` to the current timestamp before calling `set_update_timestamp`, instead of calling `set_update_timestamp` with `nil`.

Assuming that analysis is correct, this patch doesn't change the behavior, since in that case the create timestamp is not overwritten, and set_update_timestamp will set `time` to the current timestamp (it uses `time || model.dataset.current_datetime`).  If you wanted both the create and update timestamps to the same time when the create timestamp was already set, you'd have to get the existing value from the create timestamp, and set the update timestamp to that time.  That would break backwards compatibility, so it would have to be added as a separate option if that behavior was desired.

I pushed additional specs that pass both with and without this patch, showing the behavior does not change.  Since this patch doesn't change behavior, it becomes only a style change, and while the style is a little better, it hurts performance in the default case when create timestamp is already set, since your patch changes things to calculate the current timestamp in all cases, even when it would not be used.  For that reason,  I can't accept it.

In the future, if you think there is a bug and would like to fix it, first work on a spec that fails with the current code, then modify the code so the spec passes.  This ensures that the change you made has the effect you intended.

NicWagenaar:
Thanks for the detailed explanation. I'll have to investigate deeper why the inconsistency appears in my project. If I push anything in the future, I'll make sure to add appropriate specs :)

PR #1281, opened 2016-12-29
Add note about adding extensions

tpitale:
Spent some time searching for the appropriate way to add extensions (pg_enum, pg_json) to standalone migrations using the `sequel` CLI. Thought I should add some quick docs while I was here. Open to any suggestions to improve/add/modify. Or, if this exists somewhere else, please let me know. Maybe my googling skill isn't so great.

### DIFF
--- a/doc/migration.rdoc
+++ b/doc/migration.rdoc
@@ -635,3 +635,14 @@ database the migration is being run on, and operate accordingly:
       end
     end
   end
+
+== Adding Extensions
+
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch.  I think it makes sense to add documentation about how to use database extensions in migrations correctly. The example given is not the correct way of handling it, but I'll merge this and then fix it.

tpitale:
Oh good, I'm glad I opened this, then! That's what I did to get rid of an error. Happy to do it the right way instead. Thanks @jeremyevans!

Issue #1297, opened 2017-02-01
#dataset.columns! always returning an empty array


mahlonsmith:
### Complete Description of Issue

Sequel v4.42 `#dataset.columns!` is returning empty array.  v4.41 seems okay!  Breaks Sequel::Migration among other introspective things.

### Simplest Possible Self-Contained Example Showing the Bug

```ruby
db = Sequel.connect( 'postgres://...' )

# Note the '!', bypassing 4.42 cache
db[ :schema_migrations ].columns!    #=> []

# yet...
db[ :schema_migrations ].first #=> {:filename=>"20151102_initial.rb"}
```

First noticed this with Sequel::Migrator, due to:

    migration.rb:769:in `schema_dataset': Migrator table schema_migrations does not contain column filename (Sequel::Migrator::Error)`

It appears to be related to the column caching added in 4.42, but I'll need to do more digging to see specifically where the problem lies.

jeremyevans:
Please do more digging and post a self contained example, as I am unable to reproduce:

```
$ ls testm
20151102_initial.rb
$ cat testm/20151102_initial.rb
Sequel.migration{}
$ bin/sequel -m testm postgres:///?user=sequel_test
$ bin/sequel postgres:///?user=sequel_test -c "p DB[ :schema_migrations ].columns!; p DB[ :schema_migrations ].first"
[:filename]
{:filename=>"20151102_initial.rb"}
```

jeremyevans:
By any chance are you using an old version of sequel_pg (<1.6.17).  If so, that's what is causing the problem.  In general if you are updating Sequel, you should update sequel_pg at the same time.

mahlonsmith:
Ah, yep.  I was on 1.6.13.

I was trying to make a self contained example using in-mem sqlite, and not having much success, which was a good clue it was localized to PG... but yeah.  I should have considered that before the bug report!

Since sequel_pg is optional, clearly there can't be specific gem dependencies -- but I wonder if there's an opportunity somewhere to have a sanity check for a min version in sequel that offers less surprising/funky behavior when compatibility is broken.   Maybe if the sequel_pg gem embedded it's own version, and the check could be in the postgres adapter after require?  Spit ballin':

```ruby
MIN_SEQUEL_PG = '1.6.17'

# ...

# Returns an easily comparable string from a version number.
def comparable_version( ver )
    return ver.split('.').collect {|char| char.to_i }.pack('N*')
end

# ...
[...]

PR #1298, opened 2017-02-01
Raise early if the installed version of sequel_pg is too old to behave reliably

mahlonsmith:
Hello!  This pull request is in regards to #1297.

A few notes:

  * Just using `gem` instead of `require` seems the most natural way to do this, but rescuing/warning became more involved when dealing with bundler's different exception classing.  Trying to default to being unobtrusive, and adding special casing for bundler doesn't feel appealing.
  * If the environment doesn't have sequel_pg installed via gem, but it's still in the $LOAD_PATH, the behavior is "just as it was" beforehand.
  * I omitted the check for the Windows bit... the sequel_pg docs claim there isn't recent Windows support anyway, so it feels like you're on your own in a Windows environment regardless.  Not sure what feelings are surrounding that.

### DIFF
--- a/lib/sequel/adapters/postgres.rb
+++ b/lib/sequel/adapters/postgres.rb
@@ -9,6 +9,9 @@ begin
   raise LoadError unless defined?(PGconn::CONNECTION_OK)
 
   SEQUEL_POSTGRES_USES_PG = true
+  # The minimum version of the sequel_pg accelerator that
+  # is intended to work with this sequel
+  SEQUEL_PG_MINIMUM_VERSION = '1.6.17'
 rescue LoadError => e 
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch.  I think this is a reasonable approach in the short term.  In the long term, we probably want to add code to sequel_pg so that during extension initialization, it calls a method in Sequel, and if the sequel_pg version is not supported, sequel_pg will not replace any methods.

We can remove Windows support for sequel_pg.  I had to drop Windows support in sequel_pg because Windows doesn't allow calling functions defined in one dll from another dll, or at least I don't know enough to get it to work.  After 4.42.0, the old windows versions of sequel_pg would be broken, so there is no reason to try to load the code anymore.

PR #1299, opened 2017-02-01
fix: convert non-nil elements

mistoo:
This fixes boolean pg_array conversion:

>> DB.schema(:foos).last
=> [**:barray**, {:oid=>1000, :db_type=>"boolean[]", :default=>nil, :allow_null=>true, :primary_key=>false, :type=>**:boolean_array**, :ruby_default=>nil}]
>> x = Foo.new
=> #<Foo @values={}>
>> x.barray = [ false, true, false ]
=> [false, true, false]
>> x.barray
=> [nil, true, nil]

### DIFF
--- a/lib/sequel/core.rb
+++ b/lib/sequel/core.rb
@@ -204,7 +204,7 @@ module Sequel
     array.map do |i|
       if i.is_a?(Array)
         recursive_map(i, converter)
-      elsif i
+      elsif !i.nil?
         converter.call(i)
       end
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch. This fix looks correct to me.  Any chance you could add a spec for the behavior?  If not, I can do so.

jeremyevans:
Specs look good.  Thanks for your help!

PR #1300, opened 2017-02-02
"Communication link failure" can also mean disconnect

perlun:
We've had this in our own monkey patches for this class for some time now, but I felt it's better to submit it upstream. :wink:

### DIFF
--- a/lib/sequel/adapters/jdbc/as400.rb
+++ b/lib/sequel/adapters/jdbc/as400.rb
@@ -43,8 +43,8 @@ module Sequel
 
         def disconnect_error?(exception, opts)
           super ||
-            exception.message =~ /\AThe connection does not exist\./ ||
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch.  This looks good, but can you please combine the two error messages into a single regexp?

perlun:
@jeremyevans Unsure of how to do that, my regex fu is no up to speed I think. 😉 But, I changed it to `#start_with?` instead, which is what Rubocop recommends instead of \A regexps. Could you live with this approach?

jeremyevans:
@perlun `/\A(pattern1|pattern2)/`.  That's what we use in other adapters, and it scales better (think 5+ disconnect error types).

perlun:
Ok, fair points! Thanks for fixing. 👍🏻

PR #1312, opened 2017-02-26
[WIP] Use RETURNING statement for INSERT statements in Oracle

flash-gordon:
Hello @jeremyevans!

It took me a while to make some research on how support for `RETURNING` should be added to Sequel. I copied most of the code from Firebird and PostgreSQL adapters and it looks OK to me, more or less. I'm planning to add support for `UPDATE` statements, this shouldn't be hard, but I want to hear feedback from you before that.

Regarding JDBC support I did some research and found that Oracle supports `RETURNING` only for prepared statements, if I understood this correctly. Sequel's default strategy is not using prepared statements and this is why it's unclear for me how this should be handled. (See how AR's adapter executes statements for details on this https://github.com/rsim/oracle-enhanced/blob/d990f945de4d972833487b1b3364a5d013549c7f/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb#L250-L271. And of course it is different from Sequel because AR does not provide an API for returning specific columns.)

Questions:
1. What do you think about the current implementation?
2. What we should do about JDBC implementation? Most of the methods I added can be transferred to `shared/oracle` safely, but starting to use prepared statements by default can have caveats I'm not aware of. Do you know about any of them? Is it fine to override `execute` completely in an adapter so the changes won't affect other adapters?

I also have some feedback after working with the project.
1. When 1.8 support is going to be dropped? It's already outdated a bit hah. I can't install it on my machine, and I think most of native adapters requires 1.9 at least.
2. There are a lot of trailing spaces in the project. Nowadays many editors trims them on file saving, it makes commit process unnecessary hard. You either can commit all the trims in files you touched and pollute the PR with irrelevant changes or manually undo trims which obviously is quite a useless activity. Configuring an editor for a project is an option, but not every editor supports per-project configuration. It would be awesome to trim all the trailing spaces in a single commit.
3. Sequel is complex! I think moving towards a design with more separated adapters will reduce significantly. It will end up with more code duplication but will make things easy to change. While I was working on `RETURNING` I had to switch between different database in order to make this feature work for all adapters and it wasn't easy because of all these db-specific nuances. I managed to avoid any changes in base classes (with one tine exception) but in case of JDBC this won't be so easy. 
Having ad-hoc `extend`s makes it very hard to understand what's the real interface is, why this method isn't called and why that method is missing. `extend` also invalidates MRI's method cache which in order affects performance. I think it's better to use a different approach in places where `extend` is used.

/cc @solnic

### DIFF
--- a/lib/sequel/adapters/oracle.rb
+++ b/lib/sequel/adapters/oracle.rb
@@ -13,23 +13,27 @@ module Sequel
       # ORA-01012: not logged on
       # ORA-03113: end-of-file on communication channel
       # ORA-03114: not connected to ORACLE
-      CONNECTION_ERROR_CODES = [ 28, 1012, 3113, 3114 ]      
-      
+      CONNECTION_ERROR_CODES = [ 28, 1012, 3113, 3114 ]
+
       ORACLE_TYPES = {
         :blob=>lambda{|b| Sequel::SQL::Blob.new(b.read)},
         :clob=>lambda(&:read)
       }
+
+      NULL = LiteralString.new('NULL').freeze
+      NULL_RETURNING = [NULL].freeze
+      NULL_RETURNING_BINDING = [[NULL, String].freeze].freeze
 
       # Hash of conversion procs for this database.
       attr_reader :conversion_procs
 
       def connect(server)
         opts = server_opts(server)
-        if opts[:database]
-          dbname = opts[:host] ? \
-            "//#{opts[:host]}#{":#{opts[:port]}" if opts[:port]}/#{opts[:database]}" : opts[:database]
+        if opts[:database] && opts[:host]
+          port = opts[:port] ? ":#{opts[:port]}" : ""
+          dbname = "//#{opts[:host]}#{port}/#{opts[:database]}"
         else
-          dbname = opts[:host]
+          dbname = opts[:database] || opts[:host]
         end
         conn = OCI8.new(opts[:user], opts[:password], dbname, opts[:privilege])
         if prefetch_rows = opts.fetch(:prefetch_rows, 100)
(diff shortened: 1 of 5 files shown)


jeremyevans:
Thank you for your work on this.  In regards to your questions:

1) The current implementation looks mostly good.  Note that the use of prepared statements in the sense of `PREPARE` at the SQL level can never be forced, it must always remain optional. It looks like this doesn't use prepared statements in that sense, it just uses bound variables, which is OK.

   The modification of the default `Dataset#execute` is not acceptable, you should just have the code in the oracle adapter as that is the only adapter that needs it.

   If I'm reading this correctly, this is not thread safe.  It appears that it relies on the mutation of `dataset.opts[:returning]` which can be shared across threads.  You'll have to find another way to handle this that is thread safe before it could be accepted.  If I'm wrong about this, please explain how thread safety is being handled.  You should probably make sure that `dataset.opts[:returning]` is frozen (as well as each array inside of it) in order to make sure you don't have any thread safety issues.

   This patch appears to include changes unrelated to prepared statements, such as `Database#connect` in the oracle adapter.  Please separate these into separate commits/pull requests.  Likewise, please try to avoid unrelated whitespace modifications in pull requests.

   For supports_returning, if only `INSERT RETURNING` is supported, you should use `type == :insert` instead of `true`.

   For the specs, use `db.adapter_scheme == :jdbc` instead of `defined?(JRUBY_VERSION)` to check if the jdbc adapter is in use.

   As I mentioned above, in general this looks good, so with the changes proposed I should be able to merge it as long as it passes all of the adapter/integration tests (other than those specifically skipped).

2) In terms of support in `jdbc/oracle`, if there is a way to get the code shared for both the `oracle` (oci8-based) and `jdbc/oracle` adapters, it should probably go in `shared/oracle`.  Otherwise, you should have separate implementations in each adapter, specific to that adapter.  It looks like the `primary_key`, `returning_value`, and `disable_insert_returning` methods could potentially move to `shared/oracle`, for example.

In terms of your general feedback:

1) Ruby 1.8 support will be dropped in Sequel 5, which will be released later this year.

2) As mentioned in the contribution guidelines, Sequel is not pedantic about trailing whitespace.  For new code contributions, feel free to avoid trailing whitespace, but do not submit changes just to remove trailing whitespace or include whitespace changes with code commits. You'll actually find that stripping all trailing whitespace will break the specs (try it if you want).

3) I appreciate your concern about complexity, and agree that Sequel is complex.  However, I do not plan to significantly change the architecture, as I do not see the benefits of that exceeding the costs.  In the cases where Sequel potentially uses `extend` at runtime, it's usually mentioned in the documentation, and on recent versions of MRI the use of `extend` does not harm performance for other objects as it only clears the cache for the class that is modified (instead of forcing a global cache flush as earlier versions of MRI did).

   If you have specific, backwards-compatible fixes for any particular issue, I'm certainly willing to consider pull requests for them.

As mentioned in the contribution guidelines, as this is not ready for merging, you should probably ask for feedback on IRC or the Google Group, instead of submitting a pull request.  Since this is not ready for merging, I'm going to close this now.  When you think you have a patch ready for merging, please update this pull request and ask me to reopen, or submit a new pull request. If you have questions or you would like feedback/code review, please ask on IRC or the Google Group with a link to the commit/branch.

flash-gordon:
Ah, sure, thanks for the feedback! I should send a PR preview or just a patch to that GT thread, sorry about this :) I'll continue to work on this, the part about JDBC is still not completely clear so I'll dig into docs about what `prepareStatement` really is then. Other parts are clear, will ask questions in mentioned channels.

PR #1313, opened 2017-02-28
remove additional/unused variable assignment

tagomoris:
Ruby (with warning option) warns about it:
```
/Users/tagomoris/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/sequel-4.43.0/lib/sequel/model/base.rb:2045: warning: assigned but unused variable - m
```


jeremyevans:

PR #1328, opened 2017-03-23
Schema dump ignoring database schema

dadario:
Passing options from schema_dump method to schema method, so database
can differ from schemas

A database with multi schemas don't diff the schema that must be filtered.
As schema method already accept schema options, just passing from
method caller


jeremyevans:
Thanks for the patch. I think this makes sense. Can you add a spec that checks that the :schema option is actually being passed?  If not, I can do so when I merge.

PR #1329, opened 2017-03-23
Dump collate for mysql

stereobooster:
What is proper way to dump collation info? Should be done as plugin?

UPD: found this one https://github.com/coupler/sequel-collation/blob/master/lib/sequel/extensions/collation/adapters/mysql.rb


jeremyevans:
I don't think it should be done at all in core Sequel, as it invites a slippery slope.  There are many, many database-specific aspects that the schema dumper does not currently handle (e.g. functions, triggers, permissions), and trying to support them all is not feasible.  If you are using database specific features, you should probably be using the database specific dump tools.

PR #1331, opened 2017-03-29
Add type conversion for MS SQL DATETIMEOFFSET

iaddict:
MS SQL Server with a jdbc connection has a type microsoft.sql.Types.DATETIMEOFFSET which currently is not mapped to a ruby type.
This pull request adds a mapping for this type to ruby Time.


iaddict:
This was not yet ready for prime time. Sorry!

PR #1333, opened 2017-03-29
Fixed a bug of Model#[] when using the split_values plugin.

bananarne:


jeremyevans:
Thanks for the report. I agree that the current behavior has a bug in regards to handling false values.  I'll merge this shortly and add a spec for it.

PR #1334, opened 2017-04-10
Indicate in error message which is the file with the empty migration

jmettraux:
Helps pinpointing issues in migrations. At least for me.

Many thanks for Sequel.


jeremyevans:
Thanks for the patch. This looks good to me.  I'll merge and test and assuming no problems it should be pushed up to GitHub shortly.

PR #1337, opened 2017-04-12
Fix conversion for jdbc-SQLServer TIME values

iaddict:
Time values with usecs got mangled with the default RubyTime conversion method, because java.sql.Time#toString rounds usecs (eg. 12:30:00.999999 => 12:30:01).
This patch fetches the result as string which yields the desired format for conversion with Sequel.string_to_time.

### DIFF
--- a/lib/sequel/adapters/jdbc/sqlserver.rb
+++ b/lib/sequel/adapters/jdbc/sqlserver.rb
@@ -14,12 +14,29 @@ module Sequel
       end
     end
 
+    class TypeConvertor
+      def MSSQLRubyTime(r, i)
+        # MSSQL-Server TIME should be fetched as string to keep the precision intact, see:
+        # https://docs.microsoft.com/en-us/sql/t-sql/data-types/time-transact-sql#a-namebackwardcompatibilityfordownlevelclientsa-backward-compatibility-for-down-level-clients
+        if v = r.getString(i)
+          Sequel.string_to_time("#{v}")
+        end
+      end
+    end
+
     # Database and Dataset instance methods for SQLServer specific
     # support via JDBC.
     module SQLServer
       # Database instance methods for SQLServer databases accessed via JDBC.
       module DatabaseMethods
         include Sequel::JDBC::MSSQL::DatabaseMethods
+
+        def setup_type_convertor_map
+          super
+          map = @type_convertor_map
+          types = Java::JavaSQL::Types
+          map[types.const_get(:TIME)] = TypeConvertor::INSTANCE.method(:MSSQLRubyTime)
+        end
 
         # Work around a bug in SQL Server JDBC Driver 3.0, where the metadata
         # for the getColumns result set specifies an incorrect type for the
--- a/spec/adapters/mssql_spec.rb
+++ b/spec/adapters/mssql_spec.rb
@@ -68,6 +68,14 @@ describe "MSSQL" do
   it "should should support OUTER APPLY" do
     @db[:test3].outer_apply(@db[:test4].where(Sequel[:test3][:v3]=>Sequel[:test4][:v4])).select_order_map([:v3, :v4]).must_equal [[1,1], [2, nil]]
   end
+  
+  it "should handle time values with fractional seconds" do
+    t = Sequel::SQLTime.create(10, 20, 30, 999900)
+    v = @db["SELECT CAST('10:20:30.9999' AS time(7))"].get
+    v = Sequel.string_to_time(v) if v.is_a?(String)
+    pr = lambda{|x| [:hour, :min, :sec, :usec].map{|m| x.send(m)}}
+    pr[v].must_equal(pr[t])
+  end
 end
 
 # This spec is currently disabled as the SQL Server 2008 R2 Express doesn't support


jeremyevans:
Thanks for the patch. This makes sense to me, I'll merge and test shortly.

jeremyevans:
Can you please create a spec that fails with the current code?  I tried creating one, but it passes.  In my environment, both the jdbc/sqlserver and jdbc/jtds adapters return `time` column values as strings, not `SQLTime` instances, and the strings have the correct hour/min/sec/usec values if parsed by Sequel.string_to_time. 

Here's the spec I created:

```diff
diff --git a/spec/adapters/mssql_spec.rb b/spec/adapters/mssql_spec.rb
index c024073c..5e8b9ae3 100644
--- a/spec/adapters/mssql_spec.rb
+++ b/spec/adapters/mssql_spec.rb
@@ -68,6 +68,14 @@ describe "MSSQL" do
   it "should should support OUTER APPLY" do
     @db[:test3].outer_apply(@db[:test4].where(Sequel[:test3][:v3]=>Sequel[:test4][:v4])).select_order_map([:v3, :v4]).must_equal [[1,1], [2, nil]]
   end
+
+  it "should handle time values with fractional seconds" do
+    t = Sequel::SQLTime.create(10, 20, 30, 750000)
+    v = @db.get(Sequel.cast(t, 'time'))
+    v = Sequel.string_to_time(v) if v.is_a?(String)
+    pr = lambda{|x| [:hour, :min, :sec, :usec].map{|m| x.send(m)}}
+    pr[v].must_equal(pr[t])
+  end
 end

 # This spec is currently disabled as the SQL Server 2008 R2 Express doesn't support
```

iaddict:
Please find the spec on the last commit.
I also changed the implementation slightly and use raw sql, because Sequel#select already truncates the casted value.
```ruby
t.usec #=> 999900
@db.select(Sequel.cast(t, 'time(7)')).sql #=> "SELECT CAST('10:20:30.999' AS time(7))"
```
I tried to figure out where the value gets truncated, but haven't found it.

jeremyevans:
JDBC only supports miillisecond precision for time columns, but that's not really the issue here as the jdbc/sqlserver adapter returns time values as strings.

I ran your spec without your code changes and it passes for me using the tinytds, jdbc/jtds, and jdbc/sqlserver adapters, so I'm not sure your code changes actually fix things.  When trying to fix a bug, it's best to always write a failing spec first.

In terms of the default literalization using milliseconds and not microseconds, currently Sequel uses the same precision for both time and datetime columns, and MSSQL datetime only supports milliseconds (datetime2 supports microseconds, but Sequel cannot be sure it is being used).

This patch could address those issues by allowing for time precision to differ from datetime:

```diff
diff --git a/lib/sequel/adapters/shared/mssql.rb b/lib/sequel/adapters/shared/mssql.rb
index 936db41a..c56e9b9b 100644
--- a/lib/sequel/adapters/shared/mssql.rb
+++ b/lib/sequel/adapters/shared/mssql.rb
@@ -1132,7 +1132,15 @@ module Sequel
         end
       end

-      # MSSQL supports millisecond timestamp precision.
+      # MSSQL supports 100-nsec precision for time columns, but ruby by
+      # default only supports usec precision.
+      def sqltime_precision
+        6
+      end
+
+      # MSSQL supports millisecond timestamp precision for datetime columns.
+      # 100-nsec precision is supported for datetime2 columns, but Sequel does
+      # not know what the column type is when formatting values.
       def timestamp_precision
         3
       end
diff --git a/lib/sequel/dataset/sql.rb b/lib/sequel/dataset/sql.rb
index 7ba3d459..4a205eb4 100644
--- a/lib/sequel/dataset/sql.rb
+++ b/lib/sequel/dataset/sql.rb
@@ -1093,8 +1093,8 @@ module Sequel

     # Return the SQL timestamp fragment to use for the fractional time part.
     # Should start with the decimal point.  Uses 6 decimal places by default.
-    def format_timestamp_usec(usec)
-      unless (ts = timestamp_precision) == 6
+    def format_timestamp_usec(usec, ts=timestamp_precision)
+      unless ts == 6
         usec = usec/(10 ** (6 - ts))
       end
       sprintf(".%0#{ts}d", usec)
@@ -1293,7 +1293,7 @@ module Sequel

     # SQL fragment for Sequel::SQLTime, containing just the time part
     def literal_sqltime(v)
-      v.strftime("'%H:%M:%S#{format_timestamp_usec(v.usec) if supports_timestamp_usecs?}'")
+      v.strftime("'%H:%M:%S#{format_timestamp_usec(v.usec, sqltime_precision) if supports_timestamp_usecs?}'")
     end

     # Append literalization of Sequel::SQLTime to SQL string.
@@ -1549,6 +1549,12 @@ module Sequel
       String.new
     end

+    # The precision to use for SQLTime instances (time column values without dates).
+    # Defaults to timestamp_precision.
+    def sqltime_precision
+      timestamp_precision
+    end
+
     # SQL to use if this dataset uses static SQL.  Since static SQL
     # can be a PlaceholderLiteralString in addition to a String,
     # we literalize nonstrings.  If there is an append_sql for this
```

If you are just worried about sub-millisecond time rounding in the current Sequel version, you can use the round_timestamps extension, or you change the timestamp_precision for the datasets: `DB.extend_datasets{def timestamp_precision; 6 end}`

iaddict:
Thanks for the explanation concerning timestamp_precision. This should solve the cast issue.

I also ran the sequel specs without the fix and it did fail on my setup:
```
  4) Failure:
MSSQL#test_0003_should handle time values with fractional seconds [/Users/addict/Tests/SequelMSSQL/sequel/spec/adapters/mssql_spec.rb:77]:
Expected: [10, 20, 30, 999900]
  Actual: [10, 20, 31, 0]
```

I debugged the conversion and found that TIME values are converted by this method, which uses getTime:
```ruby
      # https://github.com/jeremyevans/sequel/blob/master/lib/sequel/adapters/jdbc.rb#L69
      def RubyTime(r, i)
        if v = r.getTime(i)
          Sequel.string_to_time("#{v.to_string}.#{sprintf('%03i', v.getTime.divmod(1000).last)}")
        end
      end
```

r.getTime(i).to_string then yields the rounded value.

Maybe its an issue with the SQLServer or jdbc driver versions in use.
I test with the following versions:

```
Microsoft SQL Server vNext (CTP1.4) - 14.0.405.198 (X64) 
	Mar 11 2017 01:54:12 
	Copyright (C) 2016 Microsoft Corporation. All rights reserved.
	on Linux (Ubuntu 16.04.1 LTS)
```

And the jdbc driver found here: https://www.microsoft.com/en-us/download/details.aspx?id=11774

The SQLServer is run in a docker container found here:
https://github.com/rails-sqlserver/docker-images/tree/master/mssql-server-linux-tinytds

Many thanks for your patience!

jeremyevans:
I'm using the same JDBC driver you are using, but I'm connecting to a Microsoft SQL Server 2012 Express instance running on Windows, and I'm testing on Windows:

```
jruby 9.1.1.0 (2.3.0) 2016-05-19 fe84e89 Java HotSpot(TM) 64-Bit Server VM 25.45-b02 on 1.8.0_45-b15 +jit [mswin32-x86_64]
```

I'm guessing that it is an issue with differing versions.  Since the code looks correct to me, and doesn't cause problems in my testing, it shouldn't be a problem to merge it.  Thanks for your patience and work on this.

iaddict:
Great! Also many thanks for your patience :)

PR #1338, opened 2017-04-13
Add TypeConvertor jdbc/sqlserver DateTimeOffset

iaddict:
DateTimeOffset should be returned as a ruby Time instead of the raw jdbc type.

### DIFF
--- a/lib/sequel/adapters/jdbc/sqlserver.rb
+++ b/lib/sequel/adapters/jdbc/sqlserver.rb
@@ -22,6 +22,12 @@ module Sequel
           Sequel.string_to_time("#{v}")
         end
       end
+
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch. This looks good to me.  I'm going to need to make a few changes, because this breaks on older versions of the SQL Server JDBC driver, and the spec breaks when using jdbc/jtds.  But I should be able to fix those issues fairly easily.

jeremyevans:
The spec actually breaks on many different adapters (odbc, ado, older versions of tiny_tds), in different ways on each.  I'll have to do some more testing tomorrow and see if the issues can be fixed.

PR #1343, opened 2017-04-24
Improve documentation for window functions

janko-m:
* Added an example of changing the order to descending
* Added RDoc documentation for all window function options


jeremyevans:

PR #1345, opened 2017-04-28
Fix a misprint and if/unless confusion

mvasin:


jeremyevans:

PR #1346, opened 2017-05-04
little more informative deprecation

okliv:
to be able to understand what columns exactly are duplicated


jeremyevans:

PR #1351, opened 2017-05-09
Bugfix: Handling for Oracle 11g xe schema[:db_type]

StevenCregan:
Unfortunately Oracle 11g XE has been appending 'not null' to db_type, (i.e. `number(38) not null`), causing parsing issues with the method: `column_schema_to_ruby_type(schema)` in `schema_dumper.rb`

Noticed this error while doing migrations from Oracle to Postgresql. Specifically, Number(38,0) and Date fields that weren't nullable had been converted to text fields.

I've added to the regex an optional "not null" string, `(?:\snot\snull)?`, which has resolved my issues locally, and hasn't caused conflicts with the rake spec tasks

Let me know if there's more I can do to improve this solution

### DIFF
--- a/lib/sequel/extensions/schema_dumper.rb
+++ b/lib/sequel/extensions/schema_dumper.rb
@@ -21,7 +21,7 @@ module Sequel
     # database type is not recognized, return it as a String type.
     def column_schema_to_ruby_type(schema)
       type = schema[:db_type].downcase
-      if schema[:db_type].downcase.include =~ /\snot\snull\z/
+      if schema[:db_type].downcase =~ /\snot\snull\z/
         type = type.sub(/\snot\snull\z/, '')
       end
       case type
--- a/lib/sequel/extensions/schema_dumper.rb
+++ b/lib/sequel/extensions/schema_dumper.rb
@@ -55,7 +55,7 @@ module Sequel
         {:type=>String, :size=>($1.to_i if $1), :fixed=>true}
       when /\A(?:n?varchar|character varying|bpchar|string)(?:\((\d+)\))?(?:\snot\snull)?\z/o
         {:type=>String, :size=>($1.to_i if $1)}
-      when /\A(?:small)?money\z/o
+      when /\A(?:small)?money(?:\snot\snull)\z/o
         {:type=>BigDecimal, :size=>[19,2]}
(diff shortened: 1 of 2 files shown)


jeremyevans:
I'm OK with the general idea, but I think a simpler implementation would be do the following:

```ruby
def column_schema_to_ruby_type(schema)
  type = schema[:db_type].downcase 
  if db_type == :oracle
    type = type.sub(/ not null\z/, '')  
  end
  case type
  # ...
```

Then you can leave the rest of the method alone.

Please add a spec to test for this behavior, and rebase it into a single commit.

jeremyevans:
Please only do this on oracle, as I showed in my earlier example.  Also, for the `nvarchar2`, just adding a `2?` after the existing `n?varchar` seems simpler.  With those changes and this rebased into a single commit, it should be ready for merging.  Thanks!

StevenCregan:
From my testing, schema[:db_type] returns the datatype, not the type of database. Not sure what's going on there

jeremyevans:
If you look at my code example, I'm calling the `db_type` method to get the database type symbol.

StevenCregan:
Thanks for clarifying, I'll adjust

jeremyevans:
`(2)?` in regexp creates an extra capturing group.  Please use `2?` as previously recommended.  With that fixed and this squashed to a single commit, I should be able to merge later today.

StevenCregan:
Sorry. Also, changing to reference `:db_type` has caused other tests to fail, I should not have pushed that yet

StevenCregan:
After changing to check against :db_type == :oracle, it no longer works. I'm not sure what needs to change in order to fix the behavior

jeremyevans:
`:db_type == :oracle` is always going to be false.  However, the code example I gave is bad, as the method is named `database_type`, not `db_type`.  Changing `:db_type` to `database_type` should fix things.

StevenCregan:
Using :database_type == :oracle didn't work for me either

jeremyevans:
`:database_type == :oracle` is never going to be true, as the symbols aren't the same.  You need to use `database_type` (the method), not `:database_type` (the symbol).  As a general rule, if I said it was wrong before, reverting to it is probably not a wise idea.

StevenCregan:
My sincere apologies, I really don't know Ruby and didn't quite know what I was doing. I reverted simply so I could use my temporary solution; I should have closed this instead of creating clutter.

Thanks for committing a fix though!

PR #1357, opened 2017-05-12
SchemaGenerator#column: Added error handling if invalid opts are passed in

perlun:
This would have helped me while debugging a local, user-caused issue. :wink:

### DIFF
--- a/lib/sequel/database/schema_generator.rb
+++ b/lib/sequel/database/schema_generator.rb
@@ -126,6 +126,8 @@ module Sequel
       #            creating a unique index on the column.
       # :unique_constraint_name :: The name to give the unique key constraint
       def column(name, type, opts = OPTS)
+        raise ArgumentError, "opts was expected to be a Hash, not a #{opts.class.name} with " \
+                             "value #{opts.inspect}" unless opts.is_a?(Hash)        
         columns << {:name => name, :type => type}.merge!(opts)
         if index_opts = opts[:index]
           index(name, index_opts.is_a?(Hash) ? index_opts : {})


jeremyevans:
Sequel doesn't do this type of defensive type checking in most cases.  `Hash#merge!` should give you an exception if the type is not valid.  If you really want a more descriptive error message when a non-Hash is passed to `Hash#merge!`, you should probably provide a patch to ruby itself.

perlun:
> If you really want a more descriptive error message when a non-Hash is passed to Hash#merge!, you should probably provide a patch to ruby itself.

Fair point. It is really an undefined behavior to try to merge a Hash with a non-Hash. I'm fine with not including this here, like you say it better belongs in MRI itself. Thanks anyway.

perlun:
Thinking about this a bit further: @jeremyevans, would you be willing to accept a patch that lets the `column` DSL method accept this syntax in addition to the existing one?

```ruby
db.create_table('my_table') do
  column 'id', String, :primary_key # Symbol instead of options hash
end
```

...which would be equivalent to the following (supported by the method at the moment):

```ruby
db.create_table('my_table') do
  column 'id', String, primary_key: true
end
```

The core problem was that I read the guide (a bit sloppily, I must admit) and got the impression that the first syntax would work. I don't think it would be stretching it too far to coerce a single Symbol to a `{ symname: true }` hash if that's what we get. It feels like it would be an ergonomically friendly way to let the user use the DSL.

Your thoughts?

jeremyevans:
No, the third argument to column should be an option hash.  The only cases where we accept a regular argument in place of an options hash is for backwards compatibility.  I don't want to introduce new cases.

perlun:
> No, the third argument to column should be an option hash. The only cases where we accept a regular argument in place of an options hash is for backwards compatibility. I don't want to introduce new cases.

I'm fine with that for consistency. But, then I would suggest reconsidering and making the code a bit more defensive. As a user of the library, this kind of stack trace below isn't so helpful. What merge? What type conversion? You more or less _have to read the source_ of the method to understand why it's failing like this.

That, to me as a user, feels a bit unfriendly and is the reason why I want to improve on this. Your opinion remains firm: you think the only acceptable way to approach this is to patch the MRI code and submit a code upstream?

```
TypeError: no implicit conversion of Symbol into Hash
	from /Users/per/git/sequel/lib/sequel/database/schema_generator.rb:129:in `merge!'
	from /Users/per/git/sequel/lib/sequel/database/schema_generator.rb:129:in `column'
	from (irb):6:in `block in irb_binding'
	from /Users/per/git/sequel/lib/sequel/database/schema_generator.rb:40:in `instance_eval'
	from /Users/per/git/sequel/lib/sequel/database/schema_generator.rb:40:in `initialize'
	from /Users/per/git/sequel/lib/sequel/database/schema_methods.rb:248:in `new'
	from /Users/per/git/sequel/lib/sequel/database/schema_methods.rb:248:in `create_table_generator'
	from /Users/per/git/sequel/lib/sequel/database/schema_methods.rb:212:in `create_table'
	from (irb):5
	from bin/sequel:252:in `<top (required)>'
from /Users/per/.rvm/gems/ruby-2.4.1/bin/bundle:22:in `<main>'
```

jeremyevans:
I don't think littering the codebase with defensive error checks for nicer error messages is a worthy trade off.  If the error messages are unhelpful in this case, they are probably unhelpful in all cases, and should be fixed upstream.

perlun:
> I don't think littering the codebase with defensive error checks for nicer error messages is a worthy trade off.

Fair enough. Developing software frameworks of all kinds is largely a matter of tradeoffs. Like you say, in this case, you don't consider this to be "in scope" for what is worthwhile, since it will clutter the codebase quite a lot.

I'll see if I can find some time and energy for improving this upstream.

perlun:
(Another way I thought of if we want to do something about this ourselves but still not destroy the simplicity of our code is to implement our own `Sequel#merge_hash` method or something, that will have this kind of error checking built-in. That way we can call that method instead of `Hash#merge` consistently, and get it in all places where it would be useful. It would feel like a reasonable compromise between user-friendliness and readability of the Sequel code. Would you be willing to consider such a patch for inclusion?)

jeremyevans:
No, as that still adds defensive error checking, it just centralizes it.  Less litter I suppose, but the objection still applies.  I still think you should try to upstream the improvement to the error message if you think it should be changed.

PR #1358, opened 2017-05-15
Added test for incorrect error logging in `table_exists?`

perlun:
This PR adds the missing spec mentioned in #1356. It doesn't attempt to _fix_ the problem, but it illustrates how it can easily be reproduced.

Note: I haven't really used minitest before; it might be that this can be more elegantly described in some other way. (Much more familiar with `rspec`, which has a very different world view with regards to mocking.)

(The whitespace diffs came as my editor is set to trim trailing whitespace on file save. Can strip away that part of the change if you like; I included it for now since it makes sense to me personally.)

### DIFF
--- a/spec/integration/schema_test.rb
+++ b/spec/integration/schema_test.rb
@@ -25,7 +25,7 @@ describe "Database schema parser" do
       begin
         DB.schema(:items, :reload=>true).must_be_kind_of(Array)
         DB.schema(:items, :reload=>true).first.first.must_equal :number
-      ensure 
+      ensure
       end
     end
 
@@ -42,7 +42,7 @@ describe "Database schema parser" do
       begin
         DB.schema(ds, :reload=>true).must_be_kind_of(Array)
         DB.schema(ds, :reload=>true).first.first.must_equal :number
-      ensure 
+      ensure
         DB.identifier_output_method = :reverse
         DB.identifier_input_method = :reverse
         DB.drop_table(:items)
@@ -207,7 +207,7 @@ describe "Database index parsing" do
     DB.drop_index(:items, [:n, :a])
     f.call.must_equal({})
   end
-  
+
   it "should not include a primary key index" do
     DB.create_table!(:items){primary_key :n}
     DB.indexes(:items).must_equal({})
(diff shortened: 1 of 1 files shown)


jeremyevans:
This isn't a bug, it's expected that an error would be logged in this case, since the query raises an exception. `table_exists?` is implemented by trying to select from the table, and returning false if doing so raises a `DatabaseError`. 

#1356 described a situation where the exception being raised was not a DatabaseError, which would be an bug.  That's why I asked for a failing spec, to see if it was reproducible.

If you really want to filter these error messages from your logs, you can override the logger to suppress the logging of these error messages.

As mentioned in the contribution guidelines, please do not submit whitespace changes with code changes.

perlun:
> This isn't a bug, it's expected that an error would be logged in this case, since the query raises an exception. table_exists? is implemented by trying to select from the table, and returning false if doing so raises a DatabaseError.
> 
> #1356 described a situation where the exception being raised was not a DatabaseError, which would be an bug. That's why I asked for a failing spec, to see if it was reproducible.

Yes. I looked at it a bit more now, and you're absolutely right here: the log error is printed _before_ we reach the `table_exists?` method. So the problem is not that the exception coming there is `PG::UndefinedTable`, it gets handled much earlier and uses the default behavior in Sequel (which is to log errors encountered in queries to the `logger` object). Sorry for the confusion here, my bad.

----

> If you really want to filter these error messages from your logs, you can override the logger to suppress the logging of these error messages.

I might have to do something like that. The thing is, we use an abstraction layer on top of Sequel which sets things up the way we want it by default (disabling identifier mangling for example), and it also adds a preconfigured logger. So we tend to enable logging of SQL errors by default since they are useful when debugging, but in this case it's not really an "unexpected" error but more an "expected" error...

The easiest way around is probably to set `db.logger` to `nil` when making these calls, and revert it back to the real logger afterwards. A bit of a kludge, but it'll do for now. Thanks for your help in understanding this issue better.

> As mentioned in the contribution guidelines, please do not submit whitespace changes with code changes.

Sorry for that part, I read them now and will be more careful to avoid noise like that in the future.

PR #1359, opened 2017-05-15
Added note about the sequel-jdbc-as400 gem

perlun:
I just read this message when running a Sequel-based app and realized it could make sense to change this, to avoid people doing multiple versions of the adapter when the work has already been started there.


jeremyevans:
This makes sense to me.  Thanks for the patch!

perlun:
> Thanks for the patch!

No problemo.

PR #1368, opened 2017-05-26
migration.rdoc: Added whitespace

perlun:
Just a minor change I thought about while reading the docs. What do you think?

### DIFF
--- a/doc/migration.rdoc
+++ b/doc/migration.rdoc
@@ -381,20 +381,20 @@ Sequel doesn't come with generators that create migrations for you.  However, cr
 is as simple as creating a file with the appropriate filename in your migrations directory that
 contains a <tt>Sequel.migration</tt> call.  The minimal do-nothing migration is:
 
-  Sequel.migration{}
+  Sequel.migration {}
 
 However, the migrations you write should contain an +up+ block that does something, and a +down+ block that
 reverses the changes made by the +up+ block:
 
   Sequel.migration do
-    up{}
-    down{}
+    up {}
+    down {}
   end
 
 or they should use the reversible migrations feature with a +change+ block:
 
   Sequel.migration do
-    change{}
+    change {}
   end
 
 == What to put in your migration's +down+ block


jeremyevans:
Sequel style is no space between method names and opening block brace if the method doesn't take arguments.  Most (though not all) of the codebase is like that.

perlun:
> Most (though not all) of the codebase is like that.

OK, makes sense to keep the documentation in line with the project's official style. Btw, have you considered making a `.rubocop.yml` file tweaked to follow the "official" Sequel style? Many of us are using Rubocop for our own projects, and our editors are configured to run it automatically etc. Having a project-specific `.rubocop.yml` would make it easier for everyone to follow the suggested style.

However, if you're completely against Rubocop and find it too much of a nitpick, then I will do my best to not try to push this subject any further. 😇

jeremyevans:
I don't think it makes sense to have a `.rubocop.yml` file considering that Sequel development doesn't use rubocop.  It makes sense to have one when the project uses it and enforces the use of it, but that is not the case with Sequel.

perlun:
> I don't think it makes sense to have a .rubocop.yml file considering that Sequel development doesn't use rubocop.

Hehe, that's not quite what I meant. 😄 If we were to add a Rubocop settings file, we should of course use Rubocop to some degree. (and as mentioned, some of us already use it)

If we (= you plus the rest of the community) don't want to use Rubocop, it's a different matter. Is that what you're saying (you feel that it is too picky, too much work to introduce it at this stage, don't see the need, whatever)? Note that I'm not in any way making _demands_ here, merely asking questions. I personally use Rubocop a lot and find it a very useful tool since it helps me write a lot better Ruby code than otherwise.

jeremyevans:
Automated style checkers lead to a focus on how the code looks, as opposed to how it works.  Every development tool added is a barrier to contribution. I don't feel the trade off is worth it.

perlun:
I respect your opinion on this @jeremyevans. I see it quite differently. To me, (enforced) uniformity means that all code will look the same which means we can focus _more_ on how it works instead of get distracted by variety in how it looks. (People are different; I will see how it looks whether I have a tool that reminds me about it or not. Others might not.)

But Sequel is owned by you, so I respect your right to choose in this matter.

PR #1373, opened 2017-06-05
Add sequel-combine to plugin list.

wmaciejak:


jeremyevans:
Thanks for the patch!

Issue #1377, opened 2017-06-13
PG deprecation warnings


TiteiKo:
### Complete Description of Issue

When using Sequel (4.47.0) with PG 0.21.0, deprecation warnings (about PG) are displayed.

No warning with PG 0.20.0, this was introduced by https://bitbucket.org/ged/ruby-pg/commits/3ce331d8c88ec928ad0349e47ffd717f34fa6104

### Simplest Possible Self-Contained Example Showing the Bug

```ruby
# Gemfile
source 'https://rubygems.org'

gem 'sequel'
gem 'pg'
```

```ruby
# main.rb
require 'bundler/setup'
require 'sequel'

DB = Sequel.connect('postgres://postgres/pg_warnings')
```

```
$ ruby main.rb
The PGconn, PGresult, and PGError constants are deprecated, and will be
removed as of version 1.0.

You should use PG::Connection, PG::Result, and PG::Error instead, respectively.

Called from /bundle/gems/sequel-4.47.0/lib/sequel/database/connecting.rb:111:in `require'
```

jeremyevans:
Closing as we'll be tracking completion of this on #1378.

PR #1378, opened 2017-06-13
Resolve PG deprecation warnings

### LINKED ISSUE #1376: Update 'Sequel' gem caused deprecation warnings.
I have updated sequel gem to its latest version 4.47.0, It gives deprecated method warnings like,

`SEQUEL DEPRECATION WARNING: Setting a model class dataset to an invalid dataset is deprecated and will be removed in Sequel 5.  Either use a valid dataset or set require_valid_table = false for the model class.`

`SEQUEL DEPRECATION WARNING: Sequel::Model.def_dataset_method is deprecated and will be removed in Sequel 5.  Define the method inside a dataset_module block, or use the def_dataset_method_plugin.`

orien:
Substitute `PGconn`, `PGresult`, and `PGError` for `PG::Connection`, `PG::Result`, and `PG::Error` respectively.

Fixes #1376 #1377

### DIFF
--- a/doc/opening_databases.rdoc
+++ b/doc/opening_databases.rdoc
@@ -305,7 +305,7 @@ The following additional options are supported:
                     if using the pg library).
 :driver_options :: A hash of options to pass to the underlying driver (only respected if using the pg library)
 :encoding :: Set the client_encoding to the given string
-:notice_receiver :: A proc that be called with the PGresult objects that have notice or warning messages.
+:notice_receiver :: A proc that be called with the PG::Result objects that have notice or warning messages.
                     The default notice receiver just prints the messages to stderr, but this can be used
                     to handle notice/warning messages differently.  Only respected if using the pg library).
 :sslmode :: Set to 'disable', 'allow', 'prefer', 'require' to choose how to treat SSL (only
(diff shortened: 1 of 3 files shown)


jeremyevans:
Changing the constants breaks usage of the postgres adapter with the postgres-pr driver.  It also breaks usage with pg <0.13.0.   The correct way to fix this is to assign the appropriate constant  to a constant in the Sequel::Postgres namespace (`Sequel::Postgres::PGError = defined?(PG:Error) ? PG::Error : PGError`).  That should only require 3 lines of code in the adapter, and no other code changes, since ruby's constant lookup will look for the constants in Sequel::Postgres before Object.

csuhta:
@jeremyevans What's the possibility of getting a new patch release soon that includes this change? Right now if you have `gem "sequel_pg", "~> 1.6.19", require:"sequel"` and you don't specifically request an older version of `pg`, you will indirectly bundle `pg (0.21.0)` and `sequel (4.47.0)` together which trips this deprecation warning.

jeremyevans:
Sequel releases around the first of every month, so the next release should be in a little over a week, and will include the fix for the deprecation warning.  As this is just a deprecation warning and does not cause an error, I don't plan on issuing an additional release just to work around the issue.

tpberntsen:
Thanks for fixing this. 👍

PR #1382, opened 2017-06-19
Travis: jruby-9.1.12.0

olleolleolle:
This PR updates the CI matrix to use latest JRuby.

http://jruby.org/2017/06/15/jruby-9-1-12-0.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ rvm:
   - jruby-18mode
   - jruby-19mode
   - jruby-9.0.5.0
-  - jruby-9.1.10.0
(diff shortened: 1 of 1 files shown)


jeremyevans:
This looks fine.  However, in the future, there's not really a need to bump this file for every tiny release of JRuby or MRI.  If it has been a few months since the last time it was bumped, or there was a major version released, or you think there is an issue with the newest release, then it is probably worth doing.  Otherwise, it just adds churn.

olleolleolle:
I will be more careful in my updates in the future.

PR #1384, opened 2017-06-21
Postgres JDBC adapter - code breaks inside rescue

aakashAu:
If bad params are provided to the copy_into command, copier object doesn't get instantiated. This causes an error (NoMethodError Exception: undefined method `cancelCopy' for nil:NilClass) at [line: 76](https://github.com/jeremyevans/sequel/blob/1c409a402257095cf2d0f936404f1fa25182f0fe/lib/sequel/adapters/jdbc/postgresql.rb#L76).
This masks the original error being raised.


jeremyevans:
Change looks good, thanks for the patch!

PR #1387, opened 2017-06-25
Preserve encoding when parsing PostgreSQL arrays

jeltz:
When parsing a PostgreSQL text array we want to have the same encoding
set for the members of the array as we had for the array literal, rather
than having them use Ruby's ASCII-8BIT encoding.

I encountered this issue when another gem, simpleidn, turned out to be picky about receiving ASCII-8BIT encoded input. That could be a bug in simpleidn, but I still think this should be fixed in Sequel.

### DIFF
--- a/lib/sequel/extensions/pg_array.rb
+++ b/lib/sequel/extensions/pg_array.rb
@@ -424,7 +424,7 @@ module Sequel
           super(source)
           @converter = converter 
           @stack = [[]]
-          @recorded = String.new
+          @recorded = new_entry_buffer
         end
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch. This looks OK, but you'd need to make it conditional so it didn't run on ruby 1.8.  The easiest way to do that is probably to move the string creation to a method and define a different method on ruby 1.8.

jeltz:
Cool, I have added support for 1.8 now.

jeremyevans:
Looks good.  I'll try to merge and test on Tuesday.

PR #1388, opened 2017-06-28
Feature/warning void var

berkes:
Two warnings issued when including sequel. Both seem valid concerns and are fixed in this PR.

To reproduce run `ruby -w -e "require 'sequel'"`

### DIFF
--- a/lib/sequel/model.rb
+++ b/lib/sequel/model.rb
@@ -141,7 +141,7 @@ module Sequel
       end
     end
       
-    ANONYMOUS_MODEL_CLASSES = @Model_cache # :nodoc:
+    ANONYMOUS_MODEL_CLASSES = (@Model_cache ||= {}) # :nodoc:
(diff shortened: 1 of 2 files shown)


berkes:
I was aware of https://github.com/jeremyevans/sequel/issues/1184.

However, this here points to an instance variable that is assigned to a constant before being instantiated. That at least is a codesmell, but quite probably even a bug.

jeremyevans:
I think both of these warnings should be fixed, and this pull request appears to be the best way to fix them.  Thanks for the patch, I'll merge and test later today.

berkes:
Please note that I did not manage to run the test; somehow the `bundle exec rake` failed (before and after my changes). I could not find instructions to setup&run the test-suite. 
**These changes were only tested manually on my SQL setup**

jeremyevans:
@berkes For the rake failures, you should probably post on the sequel-talk Google Group and include the full backtrace.

PR #1396, opened 2017-07-11
Separate out selecting class logic in sti plugin

Aryk:
Now you can select the subclass based on the type value.
This is useful in APIs and other places where you don't
know which subclass to instantiate.


jeremyevans:
Looks good.  I'll merge and test later today.

PR #1397, opened 2017-07-14
Add support for Oracle 12 limit/offset syntax

timon:
* Add special hanling for limit N offset 0

Oracle query optimizer can not recognize ROWNUM < 10 when ROWNUM is
aliased in subquery.

Adding special check for offset being 0 eliminates subquery, allowing
database to detect COUNT STOPKEY operation

* Use new syntax for FETCH NEXT ROWS for Oracle starting with v12, unless locking is used


jeremyevans:
Thanks for the patch. This looks good, I'll merge and test later today.

PR #1399, opened 2017-07-15
Fix typos in advanced_associations.rdoc

h8is2w8:


jeremyevans:
Thanks for the patch!

PR #1403, opened 2017-07-24
schema_methods: Added support for Bignum columns

perlun:
Before this change, this code would work:

```ruby
db = Sequel.connect('jdbc:postgresql://localhost/foo')
db.create_table(:columns_types) do
  column :d, :Bignum
end
```

...but this could would fail:

```ruby
db = Sequel.connect('jdbc:postgresql://localhost/foo')
db.create_table(:columns_types) do
  # Fails with Unsupported ruby class used as database type: Bignum
  column :d, Bignum
end
```

This was also inconsistent with the migration support, where this code would actually work (the example comes from https://github.com/jeremyevans/sequel/blob/master/doc/schema_modification.rdoc which is what got me confused in the first place):

```ruby
Sequel.migration do
  change do
    create_table(:columns_types) do
      Bignum :d                           # bigint
    end
  end
end
```

This commit adds support for the failing `Bignum` types when creating tables in a non-migration scenario as well.

### DIFF
--- a/lib/sequel/database/schema_methods.rb
+++ b/lib/sequel/database/schema_methods.rb
@@ -964,6 +964,11 @@ module Sequel
       type_literal_generic_numeric(column)
     end
 
+    # Sequel uses the bigint type by default for Bignums.
+    def type_literal_generic_bignum(column)
+      :bigint
+    end
+    
     # Sequel uses the bigint type by default for :Bignum symbol.
     def type_literal_generic_bignum_symbol(column)
       :bigint


jeremyevans:
This support was deprecated in 4.35.0 and removed in 4.41.0 because it breaks on Ruby 2.4.  On Ruby 2.4, Bignum is a deprecated alias for Integer, so using the Bignum class reference results in a 32-bit integer instead of a 64-bit integer.  The current behavior, while unfortunate, is expected and will not be changed.  If you want to support this, it's fairly easy to add via a database extension (just revert a5306af1dd89c031475583807c67ae78c4c172b4), but you are on your own in terms of support.

perlun:
For reference, [here](http://blog.bigbinary.com/2016/11/18/ruby-2-4-unifies-fixnum-and-bignum-into-integer.html) is a blog post that describes this. Thanks for reminding me, I had read about it but didn't realize this was the cause of the removal.

However, dropping it entirely is like you say unfortunate. Perhaps we can find a middle ground? I added a commit that will raise a user-friendly Error in case someone tries to use `Bignum`, instructing them how to change their code. Would that be OK from your POV?

----

[This line](https://github.com/jeremyevans/sequel/blame/master/doc/schema_modification.rdoc#L37) in the schema modification docs is what got me here in the first place. It feels odd that it actually works to use a `Bignum` in migrations, but not in the scenario described above. I am not that familiar with the Sequel codebase but I presume the migration code uses a different file for evaluating its blocks.

Should `Bignum` support be dropped from the migration support (and schema_modification.rdoc) for the sake of consistency, would that make sense or are there any good reasons for retaining it the way it is?

jeremyevans:
I only break backwards compatibility in migrations to the extent that I have to.  Supporting the `Bignum` method is possible in a backwards compatible manner because it is independent of the `Bignum` class (it now uses the `:Bignum` symbol internally).  Breaking that backwards compatibility for the purpose of consistency does not seem like a worthwhile trade to me.  Restoring `Bignum` class support doesn't make sense to me, as reintroducing behavior that changes based on what Ruby version is in use is a bad idea in my opinion.

I don't think it is worth it to reintroduce partial support for the `Bignum` class on Ruby < 2.4 just to raise a nicer error message.  The current message `Unsupported ruby class used as database type: Bignum`, which to me seems reasonable.

perlun:
> I only break backwards compatibility in migrations to the extent that I have to.

Ah, makes sense... Didn't think about the fact that it's a more critical part of the code in that sense, but what you say totally makes sense! Of course, breaking people's existing migrations very much contradicts the Semantic Versioning ideas and so forth.

> Restoring Bignum class support doesn't make sense to me, as reintroducing behavior that changes based on what Ruby version is in use is a bad idea in my opinion.

Totally agree on that one, it's a horrible idea in fact.

----

However, would you accept a documentation PR to `schema_modification.rdoc` that highlights that `Bignum` _only_ works in a `create_table` block, and not in the non-migration use case? We can then point the users to somewhere reasonable (this PR, the https://github.com/jeremyevans/sequel/commit/a5306af1dd89c031475583807c67ae78c4c172b4 commit, whatever) where they can read more about why `Bignum` has this somewhat peculiar behavior.

(note: I'm not saying it's _wrong_, I'm just saying it can be confusing for people like me who read the docs and tried to applied it in a somewhat different context, running into the error.)

jeremyevans:
Sure, I'll definitely accept an update to the schema modification guide that makes the situation more clear.  However, please postpone the update until next month, as I have some local changes to the guide related to Sequel 5 and I would prefer to avoid merge/rebase conflicts

perlun:
Sure, that's fine. Eagerly awaiting Sequel 5! 😄 Is it local-only yet, there is no public preview of the code thus far?

jeremyevans:
Correct.  4.49.0 will be released August 1st and will be the last minor release of Sequel 4, and the master branch will be updated with the Sequel 5 changes shortly after the 4.49.0 release.

PR #1405, opened 2017-08-05
Allow create PostgreSQL indexes only if not exist

DyegoCosta:


jeremyevans:
I'm not opposed to adding this, but you need to add a spec (that runs on PostgreSQL 9.5+) and update the documentation.  Ideally, you would also check whether this is supported by any of the other databases that Sequel supports, and add similar support to any of them that do.

DyegoCosta:
Won't the spec I added run on PostgreSQL 9.5+? Is there any other spec I can add to that adapter's change?

The documentation to update you mean [doc/schema_modification.rdoc](https://github.com/jeremyevans/sequel/blob/master/doc/schema_modification.rdoc#add_index)? Since the other PostgreSQL specific flags aren't mentioned there I wasn't sure if I should add it so I only updated on docs on the `schema_generator.rb` where I could find them.

I see that SQLite supports this flag since `2006-01-11 (3.3.0 alpha)`, can I submit the support for it, and possibly other databases as well, in a separated PR to keep this one short? Since the SQLite's adapter doesn't implement `index_definition_sql` I believe it's gonna be a bigger change.

jeremyevans:
Sorry, apparently I didn't see the other files in the commit.  The spec looks good, except that instead of checking for specific SQL, you should run the same command again (if no exception is raised, then it works).  For the documentation, you should probably state that it doesn't create an index if an index of the same name already exists.

If SQLite is the only other database that supports this, then adding support to just the SQLite adapter may make sense.  If there is support on more than 3 databases, I generally try to add the code to the generic support, with a method to determine if the database supports it.  Other databases to check for support: MySQL, Microsoft SQL Server, SQLAnywhere, Oracle, DB2, H2, HSQLDB, Derby.

DyegoCosta:
I updated the documentation and spec.

This is also supported by SQLite, SQLAnywhere, H2, but not by the others. Looks like adding this to generic support won't affect the content of this PR, I can update this one later or follow up with a separated PR.

jeremyevans:
The new documentation looks good. The added spec seems to check the generated SQL still, and `spec/adapter_spec.rb`was modified breaking integration/adapter specs for all non-PostgreSQL databases.

jeremyevans:
Looks like the pull request branch was forced pushed and fixed the `spec/adapter_spec.rb` issue, so the removal of the check_sqls call should be all that is needed.

DyegoCosta:
Hey sorry I submitted that by mistake. I thought the check_sqls assertion could be useful to make sure we're running the proper SQL and not just bypassing the creation of the index in some other way. Do you still want me to remove it?

jeremyevans:
Yes.  In general the adapter/integration tests should not check SQLs unless there is no other possible way to test the behavior.  Not all of the current specs follow that, but I don't want to add new specs that don't.

DyegoCosta:
All set, check_sqls call removed

jeremyevans:
Looks good, I'll merge and test tomorrow.

jeremyevans:
Commit cherry-picked as 7c32436ba971c35e6b006e0d113afcad0cd332bc

PR #1412, opened 2017-08-29
Fix typo in transaction isolation docs

thoughtless:


thoughtless:
"Using this option makes Sequel" instead of "Using this option make Sequel"

jeremyevans:

PR #1413, opened 2017-08-30
Fix documentation on Sequel::Dataset for import and multi_insert

chtitux:
For (at least) PostgreSQL, a single `INSERT` command is run, so with many records,
it may be usefull to use `slice: XX` option.

### DIFF
--- a/lib/sequel/dataset/actions.rb
+++ b/lib/sequel/dataset/actions.rb
@@ -320,6 +320,10 @@ module Sequel
     #   # INSERT INTO table (x, y) VALUES (1, 2) 
     #   # INSERT INTO table (x, y) VALUES (3, 4) 
     #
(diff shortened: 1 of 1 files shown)


jeremyevans:
First, thanks for working on a patch to improve the documentation.

The documentation for import already explains states "This method can be used to efficiently insert a large number of records into a table in a single query if the database supports it."  The documentation for multi_insert explains that it is a front end for import.

The example code you are providing showing a single query is only used by a few databases, other databases use a UNION-based approach.

Considering those two points, I don't think it makes sense to add this to the documentation.

chtitux:
Thanks for the review, I completely understand your point.

Thanks for Sequel too, that is a great piece of software used at www.trainline.eu

PR #1415, opened 2017-09-06
 Support :like option in Database#create_table on PostgreSQL, for table copy

deepj:
This allows to create a new table using copy of another table.

### DIFF
--- a/lib/sequel/adapters/shared/postgres.rb
+++ b/lib/sequel/adapters/shared/postgres.rb
@@ -892,6 +892,10 @@ module Sequel
           sql += " INHERITS (#{Array(inherits).map{|t| quote_schema_table(t)}.join(', ')})"
         end
 
+        if like = options[:like]
+          sql = sql.sub('()', "(LIKE #{quote_schema_table(like)} INCLUDING ALL)")
+        end
+
         if on_commit = options[:on_commit]
           raise(Error, "can't provide :on_commit without :temp to create_table") unless options[:temp]
           raise(Error, "unsupported on_commit option: #{on_commit.inspect}") unless ON_COMMIT.has_key?(on_commit)
--- a/lib/sequel/database/schema_methods.rb
+++ b/lib/sequel/database/schema_methods.rb
@@ -169,6 +169,8 @@ module Sequel
     #             foreign server that was specified in CREATE SERVER.
     # :inherits :: Inherit from a different table.  An array can be
     #              specified to inherit from multiple tables.
+    # :like :: Copy all collum names, data types and constraints from a different
+    #          table into another table.
     # :unlogged :: Create the table as an unlogged table.
     # :options :: The OPTIONS clause to use for foreign tables.  Should be a hash
     #             where keys are option names and values are option values.  Note
--- a/spec/adapters/postgres_spec.rb
+++ b/spec/adapters/postgres_spec.rb
@@ -72,6 +72,17 @@ describe "PostgreSQL", '#create_table' do
     @db.create_table(:unlogged_dolls, :unlogged => true){text :name}
   end
 
+  it "should create a table copying another table" do
+    begin
+      @db.create_table(:some_table) { text :name; index :name }
+      @db.create_table(:another_table, :like => :some_table)
+      @db[:another_table].columns.must_equal [:name]
+      @db.indexes(:another_table).keys.must_equal [:another_table_name_idx]
+    ensure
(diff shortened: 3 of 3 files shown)


janko-m:
From the pull request guidelines in `CONTRIBUTING`:

> 4) Do not submit whitespace changes with code changes.  Sequel is not
   pedantic about trailing whitespace, so if you have an editor that
   automatically strips trailing whitespace, you may want to turn
   that feature off.

jeremyevans:
I can't currently review this because the whitespace changes cloud out the code changes.  Please amend the commit so that it doesn't contain whitespace changes.

deepj:
@jeremyevans done

jeremyevans:
Thanks, that looks much better.

The documentation and tests make this look like it only works with temporary tables, but the PostgreSQL documentation does not indicate that LIKE is specific to or related to temporary tables (they are independent concepts).

Also, the documentation says this copies constraints, but `INCLUDING CONSTRAINTS INCLUDING INDEXES` is not used, so it only copies NOT NULL constraints.  It would probably be best to use `INCLUDING ALL`, and then in the future if someone wants to modify the like_opts, they can submit a separate patch that allow for a :like_opts option.

deepj:
@jeremyevans OK Jeremy, I'll change that. Thanks for the feedback

deepj:
@jeremyevans I made some changes. Is there any preferred way how to implement additional options for the `LIKE` clause?

jeremyevans:
You can't use `sql.sub`.  Sequel uses an append only design for SQL query building.  Any modification of an existing SQL query via sub/gsub is likely to have sharp corner cases.

Also, you shouldn't be testing for index names, you should test the index has the same columns.

For options, I would recommend `like_opts: []` for no options, and `like_opts: :indexes` and `like_opts: [:indexes, :defaults]` to pick specific options.

deepj:
@jeremyevans the problem is, there is `()` generated in https://github.com/jeremyevans/sequel/blob/master/lib/sequel/database/schema_methods.rb#L724 before `LIKE` and that causes a syntax error. So that's why I used that. Any suggestion how to make it right according Sequel API standards?

jeremyevans:
According to the PostgreSQL documentation, it appears you are putting `LIKE` in the wrong spot (https://www.postgresql.org/docs/9.6/static/sql-createtable.html).  It should be going inside the parentheses.

After trying this out more, I don't think a `:like` option to `create_table` makes sense.  It doesn't handle cases such as using LIKE more than once (which PostgreSQL supports). Instead, `LIKE` support should be implemented as a `like` method inside the PostgreSQL-specific create table generator, allowing code such as:

```ruby
  DB.create_table(:foo) do
    like :bar
    like :baz, :including=>:constraints
  end
```

deepj:
@jeremyevans I use this using SQL
```
DB.run %(CREATE TEMPORARY TABLE "#{temporary_table}" (LIKE "exchange_rates") ON COMMIT DROP)
```

In the case on my change it generates

```
CREATE TEMPORARY TABLE "temporary_table" () (LIKE "another_table")
```

And this fails. Just try it.

Good point, I missed that ... but the suggestion doesn't make sense in the case of creating temporary table. That what I started to introduce this and you could see in my initial commit the temporary table...

jeremyevans:
I've already explained the issues with supporting a `:like` option for `create_table`.  Your current approach of `sql.sub` breaks if any columns are added to the table (and yes, you can use custom columns even with `LIKE`).  Just because your patch handles your use case does not make it suitable for inclusion in Sequel.

You would need to modify the code to support:

```ruby
  DB.create_table(:temporary_table, temp: true) do
    like :another_table
  end
```

If you don't want to do that, please close this pull request and continue using raw SQL or maintain the patch separately.

jeremyevans:
Closing this now as a :like option is not going to work.  If you'd like to add a like method as mentioned above, please open a new pull request.

deepj:
@jeremyevans Jeremy, it'd be nice being more patient. It's in-progress. Already it's possible to call `create_table :another_table { like :some_table }`. What missing is more test and including/excluding like options...

jeremyevans:
@deepj Note that closing of this pull request is not a rejection of the feature, it's just a rejection of the proposed implementation (see the contribution guidelines).  When you have an implementation that adds a `like` method ready, please just add a separate pull request for it so I can review it.  I do not see a point in keeping this pull request open in the meantime.

If you really want to submit a new implementation using this existing pull request, that's OK, I can reopen this pull request in that case.

PR #1416, opened 2017-09-08
Fix buffer length passed on to writeToCopy (JRuby)

ckoenig:
Up until now the length of the original input was passed on as length of the buffer. However a call to `to_java_bytes` might return several bytes for a single character depending on the encoding. In the end only part of the actual data would have been sent to the PostgreSQL server either leading to missing line endings and thus mixing several rows into one or finding invalid bytes for certain encodings.

It would be really great to also backport this to the v4 branch. If we can help in any way, just get back to us. Thanks.

### DIFF
--- a/lib/sequel/adapters/jdbc/postgresql.rb
+++ b/lib/sequel/adapters/jdbc/postgresql.rb
@@ -65,10 +65,14 @@ module Sequel
               copier = copy_manager.copy_in(copy_into_sql(table, opts))
               if block_given?
                 while buf = yield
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch. This looks good, I'll merge and test today.

Sequel only backports fixes to the previous release (5.0.0 in this case), and only then when it is a regression in that release or it fixes a security issue.  After the release of 5.0.0, Sequel 4 should be considered no longer maintained. Since this issue has been there since the code was originally added, it is not a candidate for a backport.

PR #1418, opened 2017-09-13
Fix a warning: assigned but unused variable

JuanitoFatas:
Fix this warning:

```
~/.gem/ruby/2.4.1/gems/sequel-5.0.0/lib/sequel/model/associations.rb:2546:
warning: assigned but unused variable - res
```

I also saw this one:

```
~/.gem/ruby/2.4.1/gems/sequel-5.0.0/lib/sequel/dataset/query.rb:84: warning:
statement not reached
```

not sure how to suppress it 🤔

### DIFF
--- a/lib/sequel/model/associations.rb
+++ b/lib/sequel/model/associations.rb
@@ -2543,7 +2543,7 @@ module Sequel
 
           begin
             cbs.each do |cb|
-              res = case cb
+              case cb
               when Symbol
                 # Allow calling private methods in association callbacks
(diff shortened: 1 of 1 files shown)


jeremyevans:
This patch looks good, thanks!

You can't suppress the second warning, you have to filter it.  There's not a better way to handle that scenario, and I don't make changes that hurt performance just to avoid verbose mode warnings.

JuanitoFatas:
Thanks!

PR #1420, opened 2017-09-20
Support objects that quack like a Hash in dataset

mariuszzak:
First I'd like to thank you for your effort in providing this awesome library!

Let me tell you a short story how I came across the problem.

I'm currently using Hanami.rb framework that uses Sequel gem under the hood. 
Hanami provides some utilities for basic ruby's types. One of them is `Hanami::Utils::Hanami` class which I very often use to take advantage of helpers like `deep_symbolize!`. The instance of this class does not inherit from `Hash` class, though, nevertheless, it 'quacks' like a hash (it's implicitly castable by `to_hash` method). I was very confused when I tried to update the jsonb type column of a record via repository by passing this object mentioned above. There was an exception raised that said:

```
can't express {:foo=>"Bar"} as a SQL literal
```

I spent some time to understand what is actually wrong.  Initially, I thought it's Hanami's bug, but after digging into the issue, I think Sequel should to support objects like this following the duck typing principle, shouldn't it?

Does it make sense for you?

### DIFF
--- a/lib/sequel/dataset/sql.rb
+++ b/lib/sequel/dataset/sql.rb
@@ -80,8 +80,8 @@ module Sequel
         end
       when Integer
         sql << literal_integer(v)
-      when Hash
-        literal_hash_append(sql, v)
+      when method(:castable_to_hash?).to_proc
+        literal_hash_append(sql, v.to_hash)
       when SQL::Expression
         literal_expression_append(sql, v)
       when Float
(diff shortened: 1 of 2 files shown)


jeremyevans:
I don't think Sequel should support this.  For one, it invites the slippery slope, since then all users could say that their particular quacks-like objects should be supported.  In this case, just have the class implement `sql_literal_append`, or extend the Dataset with a module that recognizes the class you want.

PR #1423, opened 2017-09-22
Skip unnecessary database lookup of macaddr's oid

jeltz:
Since we do not parse PostgreSQL's macaddr type in the pg_inet extension
there is no reason for us to look up the oid of macaddr when specifying
a conversion proc for macaddr[], and the avoid this lookup we have to
explicitly specify a noop conversion proc.

### DIFF
--- a/lib/sequel/extensions/pg_inet.rb
+++ b/lib/sequel/extensions/pg_inet.rb
@@ -46,7 +46,7 @@ module Sequel
           if respond_to?(:register_array_type)
             register_array_type('inet', :oid=>1041, :scalar_oid=>869)
             register_array_type('cidr', :oid=>651, :scalar_oid=>650)
-            register_array_type('macaddr', :oid=>1040)
+            register_array_type('macaddr', :oid=>1040, :scalar_oid=>829)
           end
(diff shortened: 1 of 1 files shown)


jeltz:
I noticed this pointless query when debugging an issue in our application. An alternative solution I can see would be to specify the scalar_oid, but I do not see the point.

jeremyevans:
I'm OK with fixing the database lookup, since macaddr's oid is fixed.

The problem with the proposed approach is that if someone has specified an alternative conversion proc for the macaddr type, it won't be respected.  I think it may be possible to use the :scalar_oid option instead of providing a conversion proc, can you try that and see if it works?

jeltz:
It seems to work fine from my testing so pushed a new patch with that approach. Would you also like me to add a comment in the code to make clear that the conversion proc for macaddr was not left out by accident?

jeremyevans:
Thanks for the quick feedback. I don't think a comment is needed.  Let me run this through the tests right now, and if nothing breaks it will be merged.

jeltz:
Thanks for the quick replies!

PR #1428, opened 2017-10-02
Fix uninitialized warning for @server_version

nguyenmv2:
This patch fixes this warning when running Sequel on Postgres

> .rbenv/versions/2.3.4/lib/ruby/gems/2.3.0/gems/sequel-5.1.0/lib/sequel/adapters/shared/postgres.rb:527: warning: instance variable @server_version not initialized

### DIFF
--- a/lib/sequel/adapters/shared/postgres.rb
+++ b/lib/sequel/adapters/shared/postgres.rb
@@ -524,7 +524,7 @@ module Sequel
 
       # The version of the PostgreSQL server, used for determining capability.
       def server_version(server=nil)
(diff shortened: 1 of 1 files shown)


jeremyevans:
Sequel purposely does not initialize all instance variables to nil as doing so can hurt performance.  While this particular case may not hurt performance significantly, this change would invite the slippery slope.

In ruby 2.4+, you can easily filter specific warnings using the warning gem.

nguyenmv2:
Thank you for the information, I totally did not know that 👍

PR #1435, opened 2017-10-14
Expect symbols from Object#instance_variables

bjmllr:
That method stopped returning strings some time before 1.9.3, so `instance_variables.include?("@dataset")` couldn't ever be true on any interpreter currently listed in `.travis.yml`.

### DIFF
--- a/lib/sequel/model/base.rb
+++ b/lib/sequel/model/base.rb
@@ -484,7 +484,7 @@ module Sequel
           subclass.instance_variable_set(iv, sup_class_value)
         end
 
-        unless ivs.include?("@dataset")
+        unless ivs.include?(:@dataset)
           if @dataset && self != Model
             subclass.set_dataset(@dataset.clone, :inherited=>true)
           elsif (n = subclass.name) && !n.to_s.empty?
--- a/spec/model/base_spec.rb
+++ b/spec/model/base_spec.rb
@@ -381,6 +381,25 @@ describe "A model inheriting from a model" do
   end
 end
 
+describe "A model inheriting from a custom base that sets @dataset" do
+  before do
+    ::Feline = Class.new(Sequel::Model)
+    def Feline.inherited(subclass)
+      subclass.instance_variable_set(:@dataset, nil)
+      superclass.inherited(subclass)
+    end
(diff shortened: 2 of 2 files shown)


jeremyevans:
Thanks for catching this.  However, if the branch is currently always taken, and nothing currently breaks, is there a reason we wouldn't want to remove the branching with something like this (generated by `git diff -b` to avoid whitespace changes):

```diff
diff --git a/lib/sequel/model/base.rb b/lib/sequel/model/base.rb
index 8267f32dd..72ee7108b 100644
--- a/lib/sequel/model/base.rb
+++ b/lib/sequel/model/base.rb
@@ -484,7 +484,6 @@ module Sequel
           subclass.instance_variable_set(iv, sup_class_value)
         end
 
-        unless ivs.include?("@dataset")
         if @dataset && self != Model
           subclass.set_dataset(@dataset.clone, :inherited=>true)
         elsif (n = subclass.name) && !n.to_s.empty?
@@ -492,7 +491,6 @@ module Sequel
           subclass.set_dataset(subclass.implicit_table_name)
         end
       end
-      end
 
       # Returns the implicit table name for the model class, which is the demodulized,
       # underscored, pluralized name of the class.
```

bjmllr:
The old behavior is to turn off automatic dataset setting when `@dataset` already exists. This is helpful for custom model base classes that have their own conventions for setting the dataset. In the added test, `Feline` is an example of a custom base (reduced from one I'm using today) that defers setting a dataset until after the inheritance hooks are finished, by setting `@dataset` directly.

After some consideration, I think everything I liked about this behavior can be accomplished with a plugin, so I'll close this now and add a link when the plugin is ready. Feel free to merge if you want to keep the old behavior after all.

jeremyevans:
After some more thought about this, I'm going to reopen this and then merge it.  This change wasn't announced in the release notes or accompanied by a deprecation warning, and the fact that it wasn't caught because the tests weren't complete enough is not reason enough to just drop backwards compatibility.  Thanks for taking the time to explain why the previous behavior was needed.

jeremyevans:
This got merged, but apparently GitHub is having issues closing issues automatically.

PR #1441, opened 2017-10-23
Fix typo in RDocs

janko-m:
There is a double "should should", so this fixes it.


jeremyevans:
Thanks for catching this!

PR #1449, opened 2017-11-10
Add ORA-02396 as a connection error

jprice-mdsol:
ORA-02396, which is an idle timeout, should be considered a connection error.

`ORA-02396: exceeded maximum idle time, please connect again`

Existing tests pass, but there are no tests that deal with oracle connection errors.

### DIFF
--- a/lib/sequel/adapters/oracle.rb
+++ b/lib/sequel/adapters/oracle.rb
@@ -11,9 +11,10 @@ module Sequel
 
       # ORA-00028: your session has been killed
       # ORA-01012: not logged on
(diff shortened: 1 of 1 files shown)


jeremyevans:
This looks good to me.  I'll merge and test on Monday.

Is it possible to override the maximum idle time on a per connection basis?  If so, then it may be possible to add tests for this.

jprice-mdsol:
It appears to be a server side user setting that you can't override on a per connection basis. To be honest, it is really difficult to find details on Oracle errors at all.

Issue #1319, opened 2017-03-07
MySQL driver acquires a pool connection for each placeholder that needs escaping 


KJTsanaktsidis:
### Complete Description of Issue

As discussed in #784, the only way to correctly escape strings in MySQL is to do it with an open connection, since it depends on the connection's encoding and the value of `sql_mode`. Because of this, `Sequel::MySQL2#literal_string_append` uses `db.synchronize` to check a connection out of the pool.

Unfortunately, when executing a query with a large number of `?` placeholders in it using `Sequel::Dataset#fetch`, `literal_string_append` gets called once per placeholder to escape the text that must go in there. If the connection pool is under contention, this causes massive slowdown of requests; for every placeholder in a query, the thread gets sent to the back of the queue for the connection pool. We had an instance where some requests were spending ~20 seconds waiting for connections, even though the pool timeout was set to 5 seconds, for this reason.

This is not a problem if `db.synchronize`has been called at a higher level (perhaps via `db.transaction`) in the stack already, because the connection has already been checked out and it is simply a fast threadlocal access to get the connection during placeholder escaping.

I should also note that it is not strictly speaking correct to do the call to `mysql_real_escape_string` on a different connection to the one the query is going to be executed on, since the locale and sql_mode can be set on a per-connection basis.

We have therefore solved this in our use case with the following monkey patch, which wraps all of the Sequel dataset action methods in `db.synchronize`.

```ruby
module Sequel
    class Dataset
        ACTION_METHODS.each do |action|
            # Sequel does not actually have all of these methods?
            next unless self.method_defined? action
            eval_text = <<~'RUBY'
                alias_method "__old_unwrapped_%{escaped_action_name}".to_sym, "%{action}".to_sym
                def %{action}(*args)
                    if block_given?
                        db.synchronize {
                            return self.__old_unwrapped_%{escaped_action_name}(*args) { |*block_args| yield *block_args }
                        }
                    else
                        db.synchronize {
                            return self.__old_unwrapped_%{escaped_action_name}(*args)
                        }
                    end
                end
            RUBY
            eval_text = eval_text % {
                action: action,
                # Unfortunately, "__old_unwrapped_<<" is not a valid ruby method name.
                escaped_action_name: action.to_s
                                        .gsub('<', 'lt')
                                        .gsub('>', 'gt')
                                        .gsub('[', 'lbrace')
                                        .gsub(']', 'rbrace')
            }
            class_eval eval_text
        end
    end
end
```

[...]

PR #1451, opened 2017-11-22
Add PlaceholderCheckout dataset extension for MySQL

KJTsanaktsidis:
The MySQL adapter checks out a connection in order to do placeholder
escaping; this means that for queries with large numbers of placeholders
in them, a large amount of time can be spent checking out/checking in
connections while building up the SQL test.

This PR adds a dataset extension which wraps the _sql generation methods
with `db.synchronize`, so that only one connection is checked out for
the entire SQL generation process.

This should resolve jeremyevans/sequel#1319

### DIFF
--- a/lib/sequel/extensions/synchronize_sql.rb
+++ b/lib/sequel/extensions/synchronize_sql.rb
@@ -1,0 +1,48 @@
+# frozen-string-literal: true
+#
+# The synchronize_sql extension exists to work around some connection- pool
+# performance considerations in a number of adapters where escaping a string
+# as part of placeholder substitution requires an actual database connection.
+# These adapters include amalgalite, mysql2, postgres, tinytds, and JDBC
+# postgres. In these adapters, `literal_string_append` includes a call to
+# `db.synchronize` to obtain a real connection object do use for escaping the
+# passed-in string.
+#
+# This has the effect of checking out a connection from the pool for _every_
+# placeholder that needs substitution in a query. For queries with lots of
+# placeholders (e.g. IN queries with long lists), this can cause the query to
+# spend longer waiting for a connection than the actual pool timeout (since
+# every individual acquisition will take less than the timeout, but the sum of
+# all of them can be greater)
+#
+# This dataset wraps all the _sql methods with a connection acquisition, so
+# there will be no need to checkout/return connections continuously during the
+# placeholder substitution.
+#
+# While this extension might solve some problems for heavily-loaded connection
+# pools with queries containing lots of placeholders, it also creates some new
+# ones. Now _every_ query is going to call `synchronize` twice, even queries
+# that have no actual placeholders, and the total time spent holding a
+# connection is going to be higher, since it's not released during the parts
+# of query preparation that are not placeholder substitution.
+
+
+module Sequel
+    class Dataset
+        module SynchronizeSQL
+            %w(
+                insert_sql
+                select_sql
+                update_sql
+                delete_sql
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch.  I'm OK with the feature in general, I just have a few changes to the implementation:

1) The documentation only talks about MySQL, but there are many Sequel adapters that operate the same way.  The documentation should be made database agnostic.

2) The documentation should discuss the downsides of use, since many queries do not require checking out a connection to generate SQL.

3) The extension should be renamed to synchronize_sql, more in line with how it operates.

4) self.included should not be used, only (select/insert/delete/update)_sql should be overridden, and they should be normal methods in the extension module (though you can still metaprogram them using `define_method` if you want).

5) The specs should use Dataset#extension, not Database#extension, and should test the synchronize call counts both with and without the extension.

KJTsanaktsidis:
Cool, I think I can make these changes.

1) I'll browse the other adapters and see I if I can find out which ones use database connections to do escaping
2) I'm not _super_ sure what the downsides would be. Obviously it's pointless extra access to the connection pool for queries which have no parameters, but is there anything else? I suppose also the connection is held for longer in total?
3) No problem, I can rename it
4) Yup, `self.included` isn't needed if there's no need to sniff out what the `_sql` methods are called - I wasn't sure if that list might change but it seems not
5) Yup I'll have another swing at the tests

KJTsanaktsidis:
I think I've addressed most of that feedback now. The only thing I'd like to raise is I also put the synchronize around `with_sql` - we have quite a few uses of just raw `db.fetch` in our code, and it seems fundamentally to suffer from the same problems with `literal_string_append`

jeremyevans:
These changes look good.  I think the only thing that still needs to be addressed is the specs.  The existing specs avoid using `let` and try to avoid using unnecessary nested contexts.  I don't think checking the number of calls to `synchronize` makes sense, since it doesn't really show the purpose of the the extension (the extension will always increase it by 1).  It should probably check when connections are checked out:

```ruby
before do
  @db = Sequel.mock
  @db.extend_datasets do
    def literal_string_append(*)
      db.synchronize{super}
    end
  end
  def (@db.pool).release(_)
     @db.checkout
     super
  end
  def @db.checkout
     @checkouts ||= 0
     @checkouts += 1
  end
  def @db.checkouts
    count = @checkouts
    @checkouts = 0
    count
  end
  @ds = @db[:tab]
end

it "should checkout connection once for insert_sql" do
  @ds.insert_sql(:foo => 8)
  @db.checkouts.must_equal(0)
  @ds.insert_sql(:foo => '8', :bar => 'barf')
  @db.checkouts.must_equal(2)

  @ds = @ds.extension(:synchronize_sql)
  @ds.insert_sql(:foo => 8)
  @db.checkouts.must_equal(1)
  @ds.insert_sql(:foo => '8', :bar => 'barf')
  @db.checkouts.must_equal(1)
end
```

I think this is better as it allows you to better see the effect of the extension.

I should have time to merge and test this on Monday.

KJTsanaktsidis:
Cool, I've gone and updated the tests to look at actual pool checkouts. I actually instrumented `assign_connection` instead of `release` because it felt a little more natural to me to measure how many times connections are _remove_ from the pool. Also, it turns out I don't need `with_sql` to be wrapped - actual placeholder substitution happens as part of one of the other `*_sql` methods after calling `fetch`. Thanks for the tip with the tests - I only noticed this because I was getting an extra checkout I wasn't expecting :)

PR #1452, opened 2017-11-25
Fix syntax error in validation_helpers documentation

janko-m:


jeremyevans:
Thanks for the patch!

PR #1456, opened 2017-12-08
Cache calls to Dataset#nullify.

chanks:
I use Dataset#nullify a fair amount in an app (it winds up being very helpful for resolving GraphQL requests, where the pagination logic is often very removed from the filtering logic), and it'd be nice if it were be able to chain cached datasets off of it.


jeremyevans:
Thanks for the patch! This looks good, I'll test and merge shortly.

PR #1459, opened 2017-12-14
allow sub-class columns to be ignored when using CTI

brianphillips:
Our use case is that we have a convention that _every_ table has an `updated_at` (or similar) column to make incremental ETLs to a data warehouse simpler. This breaks Sequel's CTI requirement that all columns in the inheritance tree be unique (excluding primary keys, of course).

The approach taken here is to have a list of columns that are ignored on _all_ sub-classes of the super class (where the `:class_table_inheritance` plugin is initialized). These columns are ignored during the initialization of the CTI plugin but also when retrieving data from the DB.


jeremyevans:
Thanks for the patch. This approach looks good to me, I'll test it and assuming no problems it should be merged today.

PR #1465, opened 2018-01-07
Add cockroachdb adapter.

hollinwilkins:
This is a first pass at a cockroachdb adapter. I noticed while making this that two main things needed to change:

1. CockroachDB only supports one savepoint, instead of trying to build this logic into Sequel, I disabled them for now

2. The timestamp format is slightly different that cockroach db expects from postgres

### DIFF
--- a/lib/sequel/adapters/cockroach.rb
+++ b/lib/sequel/adapters/cockroach.rb
@@ -1,0 +1,763 @@
+# frozen-string-literal: true
+
+require 'sequel/adapters/shared/postgres'
+
+begin 
+  require 'pg' 
+
+  Sequel::Postgres::PGError = PG::Error if defined?(PG::Error)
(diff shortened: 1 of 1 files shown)


jeremyevans:
At this point, I'm not looking to include support for cockroachdb in Sequel itself.  Please ship this as an external adapter for users who may be interested it.  After you have published the sequel-cockroachdb gem, please send a pull request to add it to the list of external adapters on the website.

I think an better approach for implementing this adapter would be requiring the postgres adapter, and making the cockroachdb adapter use subclasses of the postgres adapter classes, only overriding what needs to be changed.  That would reduce duplication of code a lot.  But making that change is up to you.

hollinwilkins:
@jeremyevans Sure thing. Will make the gem and post a comment here afterwards.

PR #1469, opened 2018-01-20
Add pretty_print plugin

tycooon:
Hi, I made a simple plugin that improves models introspection in command line tools that use `pretty_inspect`. The code is ported from ActiveRecord, because their implementation looks nice in the console.
Please take a look.


jeremyevans:
Sorry, but I don't want to ship this plugin with Sequel.  You can ship this as an external gem, and send a pull request that adds a link to it to `www/pages/plugins.html.erb`.

PR #1472, opened 2018-01-29
Fix quotes mismatch in readme

evheny0:


jeremyevans:

PR #1475, opened 2018-02-09
Fix typo in spec description

v-kolesnikov:
I was reading tests and came across to this  minor typo.


jeremyevans:

PR #1478, opened 2018-02-16
Shadowing outer local variable v

utilum:
Ruby 2.5 warns about this.


jeremyevans:
Thanks for the report. This definitely should be fixed, and I noticed some other similar issues that should be fixed as well.  I'll work on this right away.

utilum:
Thank _you_ for the quick and positive response, and for the gem, of course. Pleasure :)

PR #1479, opened 2018-02-21
Link directly to Dataset Basics guide

Erin007:
I thought this would be handy since the previous reference wasn't working for me.

### DIFF
--- a/doc/querying.rdoc
+++ b/doc/querying.rdoc
@@ -25,7 +25,7 @@ method you can use.
 === Sequel::Dataset
 
 If you are new to Sequel and aren't familiar with Sequel, you should probably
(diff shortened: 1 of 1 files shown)


jeremyevans:
Sorry, but this change is wrong.  `rdoc-ref` is the correct markup, see http://docs.seattlerb.org/rdoc/RDoc/Markup.html#class-RDoc::Markup-label-Links

See Sequel's website (or generate the documentation with rdoc yourself) to check that the links work: http://sequel.jeremyevans.net/rdoc/files/doc/querying_rdoc.html#label-Sequel-3A-3ADataset

GitHub doesn't handle `rdoc-ref` properly, but that's a bug/limitation in GitHub.  GitHub is the not the canonical source for Sequel's documentation.

PR #1486, opened 2018-03-30
Add development dependency for warning gem

timcraft:
Running the specs with warnings e.g. `rake spec_core_w` requires the warning gem, which isn't specified as a development dependency in the gemspec.


jeremyevans:
Thanks for the patch, but I don't think this is a good idea.  Only the *_w rake tasks use the warning gem.  We don't have all database drivers Sequel can use as development dependencies, even though the related adapter specs use them.  Additionally, adding warning as a development dependency would break on ruby 1.9.2-2.3, since warning requires ruby 2.4.

timcraft:
@jeremyevans Makes sense, thanks.

PR #1487, opened 2018-03-30
Fix warning: instance variable @transaction_mode not initialized

timcraft:
Fixes `warning: instance variable @transaction_mode not initialized`

This warning doesn't show up in the specs at the moment because of the way the relevant specs are setup/structured, but it can be shown by running the following Ruby script with warnings enabled:

```ruby
$:.unshift('lib')
require 'sequel'
Sequel.sqlite.transaction {}
```

### DIFF
--- a/lib/sequel/adapters/shared/sqlite.rb
+++ b/lib/sequel/adapters/shared/sqlite.rb
@@ -243,7 +243,7 @@ module Sequel
       end
 
       def begin_new_transaction(conn, opts)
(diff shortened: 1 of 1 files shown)


jeremyevans:
Sequel purposely does not initialize all instance variables to nil or call methods to access instance variables as doing so can hurt performance. While this particular case may not hurt performance significantly, this change would invite the slippery slope.  See previous issues and sequel-talk posts related to this, as well as https://bugs.ruby-lang.org/issues/10396.

PR #1489, opened 2018-04-03
add add_unique_constraint to list of reversible methods

TSMMark:


jeremyevans:
I didn't think add_unique_constraint was reversible.  It's not one of the listed reversible methods in the code. Have you tried migrating up and down using a change migration that uses `add_unique_constraint`?

TSMMark:
Hey Jeremy, you're right, it's not. I got some misinformation from a coworker and should have verified it first. My bad!

PR #1495, opened 2018-04-11
Add `rename_enum` method for `pg_enum` extension

AlexWayfer:


jeremyevans:
Thanks for the patch. These changes look good, but can you also add a spec to `specs/adapters/postgres.rb`?

celsworth:
Is there any chance this could be made reversible too? Seems like all the information (old name, new name) will be present.

AlexWayfer:
> These changes look good, but can you also add a spec to specs/adapters/postgres.rb?

Sure, I didn't think this is necessary for new method. Added.

> Is there any chance this could be made reversible too? Seems like all the information (old name, new name) will be present.

Good point! I added this, thank you.

PR #1496, opened 2018-04-14
Fix code example indentation in Association Options section

celsworth:
Noticed a tiny formatting error when reading through this doc today.


jeremyevans:
Thanks for the patch!

PR #1497, opened 2018-04-14
Ruby 2.6 Error when DEFAULT_PROXY_TO_DATASET sends :filter to an array

utilum:
2.6 introduces `Enumerable#filter` which is the same as `:select`. https://github.com/ruby/ruby/commit/b1a8c64483b5ba5e4a391aa68234e7bde6355034

Here's how I've stumbled thisupon:
```
sequel]$ ruby -v
ruby 2.6.0dev (2018-04-04 trunk 63085) [x86_64-linux]

]$ ruby spec/extensions/association_proxies_spec.rb 
Run options: --seed 1426

# Running:

EEE...

Finished in 0.009946s, 603.2618 runs/s, 2815.2216 assertions/s.

  1) Error:
Sequel::Plugins::AssociationProxies#test_0002_should send method calls to the association dataset if sent a non-array method:
ArgumentError: wrong number of arguments (given 1, expected 0)
    /$HOME/code/sequel/lib/sequel/plugins/association_proxies.rb:87:in `filter'
    /$HOME/code/sequel/lib/sequel/plugins/association_proxies.rb:87:in `public_send'
    /$HOME/code/sequel/lib/sequel/plugins/association_proxies.rb:87:in `method_missing'
    spec/extensions/association_proxies_spec.rb:28:in `block (2 levels) in <main>'

  2) Error:
Sequel::Plugins::AssociationProxies#test_0004_should reload the cached association if sent an array method and the reload flag was given:
ArgumentError: wrong number of arguments (given 1, expected 0)
    /$HOME/code/sequel/lib/sequel/plugins/association_proxies.rb:87:in `filter'
    /$HOME/code/sequel/lib/sequel/plugins/association_proxies.rb:87:in `public_send'
    /$HOME/code/sequel/lib/sequel/plugins/association_proxies.rb:87:in `method_missing'
    spec/extensions/association_proxies_spec.rb:66:in `block (2 levels) in <main>'

  3) Error:
Sequel::Plugins::AssociationProxies#test_0006_should work correctly in subclasses:
ArgumentError: wrong number of arguments (given 1, expected 0)
    /$HOME/code/sequel/lib/sequel/plugins/association_proxies.rb:87:in `filter'
    /$HOME/code/sequel/lib/sequel/plugins/association_proxies.rb:87:in `public_send'
    /$HOME/code/sequel/lib/sequel/plugins/association_proxies.rb:87:in `method_missing'
    spec/extensions/association_proxies_spec.rb:83:in `block (2 levels) in <main>'

6 runs, 28 assertions, 0 failures, 3 errors, 0 skips
```
I imagine you'd have a cleaner solution, but was happy to try.

Does this change in Ruby also raise the breaking-change notion to dispose of `:filter` as an alias of `where`? I'd be happy to try that too ;)
If so,

### DIFF
--- a/lib/sequel/plugins/association_proxies.rb
+++ b/lib/sequel/plugins/association_proxies.rb
@@ -63,9 +63,15 @@ module Sequel
       class AssociationProxy < BasicObject
         array = []
 
-        # Default proc used to determine whether to sent the method to the dataset.
+        # Default proc used to determine whether to send the method to the dataset.
         # If the array would respond to it, sends it to the array instead of the dataset.
-        DEFAULT_PROXY_TO_DATASET = proc{|opts| !array.respond_to?(opts[:method])}
+        DEFAULT_PROXY_TO_DATASET = proc do |opts|
+          array_method = array.respond_to?(opts[:method])
+          if !array_method && opts[:method] == :filter
+            warn "The behavior of the :filter method will change in Ruby 2.6. Switch to calling :where to conserve current behavior."
+          end
+          !array_method
+        end
 
         # Set the association reflection to use, and whether the association should be
         # reloaded if an array method is called.


jeremyevans:
Thank you for the report. This isn't a bug.  It's true that things break in an upgrade to ruby 2.6, but that's because ruby 2.6 is not backwards compatible with previous versions of ruby in this case.  This patch leads us down a slippery slope, where Sequel tries to undo changes in newer ruby versions to preserve the behavior in previous ruby versions.  I don't want that.  Sequel should respect the behavior of the current ruby version in use.

This patch would make the code not match the documentation.  The documented behavior is that if an array responds to the method, then it is sent to the array, otherwise it is sent to the dataset.  Your patch breaks backwards compatibility for anyone already defining `Array#filter` and currently relying on `filter` on the association proxy delegating to the array.  

The simple fact is, when you change your ruby version, if the new ruby version changes behavior, sometimes your application needs to adapt.  This has been the case before, such as when `Bignum` was made an alias to `Integer` in ruby 2.4.

To prevent surprises on upgrades to ruby 2.6, we could add a warning on ruby <2.6 if filter is used and the array doesn't respond to it, since we know that behavior will change.  However, if you run on ruby 2.6, filter should delegate to the array.  Could you please modify the patch to issue a warning instead of changing behavior?  If you don't want to do that, I can take care of it before the next release.

utilum:
Thank you, changed.
Might I also replace `filter` with `where` in the tests?

utilum:
And then... is the plan to deprecate `.filter` ?

jeremyevans:
`filter` cannot be removed until Sequel 6, and I'm not sure I want to remove it even then (in early Sequel, `filter` was preferred over `where`).  There are other Dataset methods that have different behavior from the Enumerable methods of the same name, so it's not like `filter` would be the only such method.

I'll try to test and merge this in the next few days (currently at a conference).

jeremyevans:
Cherry-picked as 5650dab5a59f0dd39be9b009690012c1f4c89f5c.

PR #1500, opened 2018-04-20
add ruby syntax highlighting

llighterr:


jeremyevans:
This is already formatted using ruby syntax highlighting: http://sequel.jeremyevans.net/rdoc/files/README_rdoc.html#label-A+Short+Example

Can you please let me know why you think this is necessary?

tycooon:
This actually doesn't work: https://github.com/llighterr/sequel/blob/26c40549a0f1d45e91324eddbe730d328b3200b5/README.rdoc

utilum:
@tycooon The only difference I see on @llighterr 's  branch is not what the PR seems to aim for.

screen shot 2018-04-20 at 17 26 53

tycooon:
Right, that's what I mean 🤦‍♂️

jeremyevans:
Since this doesn't seem to be correct or necessary, closing.

janko-m:
To clarify, the README is in the `.rdoc` format, which doesn't recognize backticks like the `.md` format, and AFAIK doesn't have a way to specify the language. The `rdoc` tool will by default syntax-higlight it in Ruby (Jeremy's first link), but GitHub will apply no syntax highlighting. This is a limitation in GitHub, it should automatically syntax-highlight RDoc code blocks as Ruby, as it is a "**R**uby Doc " format after all.

PR #1501, opened 2018-04-22
typo fix: replace 'easier that' with 'easier than'

utilum:


jeremyevans:
Thanks for the patch!

PR #1502, opened 2018-04-22
remove unnecessary line, freeze already returns receiver

utilum:


jeremyevans:
Thanks for the patch!

PR #1503, opened 2018-04-24
[WIP] Connection expiration random delay

hex2a:
This PR adds a configurable random delay (per-connection) to the connection timeout.

most tests are missing, please let me know if you have any input on that.

### DIFF
--- a/lib/sequel/extensions/connection_expiration.rb
+++ b/lib/sequel/extensions/connection_expiration.rb
@@ -37,13 +37,18 @@ module Sequel
     # connection creation before expiring a connection.
     # Defaults to 14400 seconds (4 hours).
     attr_accessor :connection_expiration_timeout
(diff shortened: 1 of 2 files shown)


jeremyevans:
This looks good, thanks for the patch!  I'm guessing this is to avoid the thundering herd if you are preallocating connections at startup. There are some thread-safety issues in the implementation, and I'll probably change this to storing an array in `@connection_expiration_timestamps` instead of adding another instance variable, but I should be able to merge and test this today.

mschfh:
thank you,
is there a date when this will be released?

hex2a:
thank you,
is there an ETA for a release?

jeremyevans:
Standard Sequel release policy is a new minor version around the first of every month.  So the next release should be around next Tuesday.

fertobar:
great! this new feature will prevent massive expiration and reconnection near same time when we use in high scalability with Puma

PR #1508, opened 2018-05-17
Add `Sequel::Model.dataset?` method

AlexWayfer:
Because `.dataset` raises error if there is no dataset,
and there is no quiet way to check dataset existing, except `rescue`.

Useful for multiple-inherited abstract classes.

Reference: https://groups.google.com/d/msg/sequel-talk/OG5ti9JAJIE/G8ncUsT4BAAJ


jeremyevans:
I think this idea is good, but please rename the method `has_dataset?`.

PR #1511, opened 2018-05-19
Fixed typo to setup rspec with sequel

hahmed:


jeremyevans:
That's not a typo, most of the documentation uses DB to represent the Sequel::Database constant.  If you are only using core Sequel and not Sequel::Model, that breaks.  I do see that the testing doc is internally inconsistent in that it uses Sequel::Model.db in other places.  I'll fix that.

hahmed:
Thanks for clarifying!

PR #1512, opened 2018-05-21
Add ability to have Sequel not log errors

stefansedich:
Currently when a logger is configured and an error occurs it will log the error and then also re-raise.

For me I want to handle these errors at a higher level and not have Sequel log the error also, this change adds the ability to turn off error logging which will now only re-raise the error where it is handled by my application.

The default behavior has not changed and it will continue to log errors if a logger is configured.

### DIFF
--- a/lib/sequel/database/logging.rb
+++ b/lib/sequel/database/logging.rb
@@ -11,6 +11,9 @@ module Sequel
     # level instead of info level.
     attr_accessor :log_warn_duration
 
+    # Whether to log errors or not
+    attr_accessor :log_errors
+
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the submitting a pull request. Unfortunately, I don't think I want to support this approach in Sequel.  Please use a custom logger instance that ignores calls to #error if you want to ignore errors.

jeremyevans:
The point is to log the SQL used, as that is often needed to debug the error.  If you don't want to use a custom logger, then: `def DB.log_exception(e, sql); end` where `DB` is your Sequel::Database instance.

PR #1517, opened 2018-05-30
Generated columns support

wjordan:
Generated Columns feature is [`T175`](https://ronsavage.github.io/SQL/sql-2003-noncore-features.html) in the SQL-2003 standard.
Supported by [MariaDB 5.2.0+](https://mariadb.com/kb/en/library/generated-columns/) and [MySQL 5.7.6+](https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-6.html).

The standard BNF grammar for a [`<generation clause>`](https://ronsavage.github.io/SQL/sql-2003-2.bnf.html#generation%20clause) is
`GENERATED ALWAYS AS (`[`<value expression>`](https://ronsavage.github.io/SQL/sql-2003-2.bnf.html#value%20expression)`)`.

This PR adds:
- `:generated_always_as` option to column definition to add a `GENERATED ALWAYS AS column expression` to column definition statements;
- `#supports_generated_columns?` method that specifies whether the database supports generated columns based on the database server version;
- `:generated` key on the schema-hash parser for the MySQL adapter.

MySQL's [`CREATE TABLE` syntax](https://dev.mysql.com/doc/refman/5.7/en/create-table.html) makes the `GENERATED ALWAYS` part optional for brevity e.g., `AS (expr)`, but it also supports the standard syntax implemented in this PR.

### DIFF
--- a/lib/sequel/adapters/shared/mysql.rb
+++ b/lib/sequel/adapters/shared/mysql.rb
@@ -19,7 +19,7 @@ module Sequel
       include Sequel::Database::SplitAlterTable
 
       CAST_TYPES = {String=>:CHAR, Integer=>:SIGNED, Time=>:DATETIME, DateTime=>:DATETIME, Numeric=>:DECIMAL, BigDecimal=>:DECIMAL, File=>:BINARY}.freeze
-      COLUMN_DEFINITION_ORDER = [:collate, :null, :default, :unique, :primary_key, :auto_increment, :references].freeze
+      COLUMN_DEFINITION_ORDER = [:collate, :null, :default, :generation, :unique, :primary_key, :auto_increment, :references].freeze
       
       # Set the default charset used for CREATE TABLE.  You can pass the
       # :charset option to create_table to override this setting.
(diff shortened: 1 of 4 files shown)


jeremyevans:
Thanks for the patch! This looks pretty good, but since MySQL is the only database with support, can you move the changes to the default Database support to just the shared MySQL adapter?

wjordan:
The feature is part of the SQL-2003 standard and isn't MySQL specific, and there are other databases other than MySQL that also support this feature (e.g., Oracle 'Virtual Columns' and SQL Server 'Computed Columns'), which is why I thought the implementation should be located in the general Database file. I can move the changes to MySQL adapter until other adapter implementations that need to share this code ever get added though.

jeremyevans:
I think that would be best, thanks!

wjordan:
All updated, please take another look.

jeremyevans:
Thanks!  Looks pretty good, I'll test soon and probably merge with some small changes.

PR #1518, opened 2018-05-31
Work around an RDoc issue

janko-m:
For some reason Window and Wrapper classes appear as top-level constants in the RDocs, without the Sequel::SQL:: prefix. I'm guessing that's because the metaprogramming that the VirtualRow class has somehow messes up the RDocs parser. This also makes the {SQL::Window} link not work in the Sequel::Dataset#over RDocs.

Since all classes before the VirtualRow class are correctly RDoc'd, and classes defined after VirtualRow outside of the Sequel::SQL namespace as well, I think defining the Window and Wrapper classes before VirtualRow should fix things.

### DIFF
--- a/lib/sequel/sql.rb
+++ b/lib/sequel/sql.rb
@@ -1777,6 +1777,61 @@ module Sequel
       end
     end
 
+    # A +Window+ is part of a window function specifying the window over which a window function operates.
+    #
+    #   Sequel::SQL::Window.new(partition: :col1)
+    #   # (PARTITION BY col1)
+    #   Sequel::SQL::Window.new(partition: [:col2, :col3])
+    #   # (PARTITION BY col2, col3)
+    #
+    #   Sequel::SQL::Window.new(order: :col4)
+    #   # (ORDER BY col4)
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the report. Can you please test and see if this also occurs with the latest release of rdoc (6.0.4)?

If we do want to work around this issue in Sequel, it would probably be best to modify the metaprogramming in VirtualRow so that it doesn't trigger the rdoc bug.  Ideally this should be fixed in rdoc, so hopefully we can put together a bug report to rdoc if the problem still exists in the latest version.

janko-m:
This still happens on RDoc 6.0.4. I also verified that this PR fixes the issue. The {SQL::Window} link not working appears to be unrelated, I'll make a separate PR for that.

> If we do want to work around this issue in Sequel, it would probably be best to modify the metaprogramming in VirtualRow so that it doesn't trigger the rdoc bug. Ideally this should be fixed in rdoc, so hopefully we can put together a bug report to rdoc if the problem still exists in the latest version.

Yeah, that makes sense. This was just the easiest workaround, because I didn't feel like going either one of these routes 😛

PR #1519, opened 2018-06-01
Fix Sequel::SQL::Window RDoc links

janko-m:
It seems that `{SomeClass}` isn't the right way to link classes in RDoc, because it doesn't work (probably mixed it up with YARD). So I changed it to the full class name, and that works. In that case the link text will be the full constant name (`Sequel::SQL::Window`).


jeremyevans:
Thanks for the patch! I'll try to merge this tomorrow before the 5.9.0 release.

PR #1520, opened 2018-06-04
Replace sequel_pg's array parser by pg's array parser

larskanis:
Caused by the [discussion on ruby-lang](https://bugs.ruby-lang.org/issues/14819) I had a look at the sequal_pg gem and compared it with what we have in pg already. I think sequal_pg could be fully integrated into pg, although it would require some efforts.

The benefit is, that we could combine the best pieces of both extensions together, so that pg could better fit the needs of all ORM libraries. In the long run this could also save some maintenance costs.

The attached patch is a first low hanging fruit, since the array parsers are pretty equal. They currently differ in that pg's parser doesn't raise any errors, but this [behavior is deprecated](https://github.com/ged/ruby-pg/commit/f0a0a2d09d0c0f82fa69a7a43643b34f34047ee7) anyways. This is the reason, why [one spec fails](https://travis-ci.org/larskanis/sequel/jobs/387825926). I could make this behavior configurable, if this is desired.

For now and for the attached patch the only benefits are:
1. The array parser could be removed from sequel_pg
2. Parsing of native types such as `int[]` is faster, since they are cased directly in C, not going through the ruby based converter.

PG has a very [flexible type cast system](https://github.com/ged/ruby-pg/#type-casts), which allows to assign and combine decoders and encoders by OID or columns based. It would be possible to add new specific decoders into PG or make PG's decoders configurable by parameters. One such example is to pass the class to be called for time-only values rather than hardcoding into the C code:
```
deco = PG::TextDecoder::Time.new(class: Sequel::SQLTime)
deco.decode("2018-06-04")
```

Is this a direction you would like to go? Or better keep sequel's `PGresult` handling separately?

### DIFF
--- a/.travis.gemfile
+++ b/.travis.gemfile
@@ -36,7 +36,7 @@ platforms :ruby do
   if RUBY_VERSION < '2.0.0'
     gem "pg", '<0.19.0'
   else
-    gem "pg"
+    gem "pg", git: "https://github.com/ged/ruby-pg/"
   end
 end
 
--- a/lib/sequel/extensions/pg_array.rb
+++ b/lib/sequel/extensions/pg_array.rb
@@ -71,6 +71,7 @@
 
 require 'delegate'
 require 'strscan'
+require 'pg'
 
 module Sequel
   module Postgres
@@ -397,23 +398,55 @@ module Sequel
         # The database type to set on the PGArray instances returned.
         attr_reader :type
 
-        # Set the type and optional converter callable that will be used.
-        def initialize(type, converter=nil)
-          @type = type
-          @converter = converter
-        end
-
-        if Sequel::Postgres.respond_to?(:parse_pg_array)
+        # Use pg's C-based parser if it has already been defined.
+        if defined?(PG::TextDecoder::Array)
         # :nocov:
-          # Use sequel_pg's C-based parser if it has already been defined.
+          class Converter < PG::SimpleDecoder
+            def initialize(converter)
+              @converter = converter
+            end
+            def decode(value, tuple=nil, field=nil)
+              @converter.call(value)
+            end
+          end
+
+          NATIVE_DECODERS = {
+            "smallint" => PG::TextDecoder::Integer,
+            "integer" => PG::TextDecoder::Integer,
+            "bigint" => PG::TextDecoder::Integer,
+            "boolean" => PG::TextDecoder::Boolean,
+            "real" => PG::TextDecoder::Float,
+            "double precision" => PG::TextDecoder::Float,
+          }
+
+          # Set the type and optional converter callable that will be used.
+          def initialize(type, converter=nil)
+            @type = type
+            params = {}
+            if converter
+              scalar = NATIVE_DECODERS[type]
+              if scalar
+                params[:elements_type] = scalar.new
+              else
+                params[:elements_type] = Converter.new(converter)
+              end
+            end
+            @deco = PG::TextDecoder::Array.new(params)
+          end
+
           def call(string)
-            PGArray.new(Sequel::Postgres.parse_pg_array(string, @converter), @type)
+            PGArray.new(@deco.decode(string), @type)
           end
         # :nocov:
         else
           # Parse the string using Parser with the appropriate
           # converter, and return a PGArray with the appropriate database
           # type.
+          def initialize(type, converter=nil)
+            @type = type
+            @converter = converter
+          end
+
           def call(string)
             PGArray.new(Parser.new(string, @converter).parse, @type)
           end


larskanis:
To be honest: This PR wasn't really meant to be merged. I'm completely new to sequel and just used a hour of spare time to hack this together. It's more about the question, whether you want to migrate `sequel_pg.gem` specifics into `pg.gem` in the long term?

Nevertheless I did [some benchmarks](https://gist.github.com/larskanis/569231d055d8dae03bed1d9a5deca6cb):

sequel-5.8.0 without sequel_pg:
```
   sequel_pluck_iary      1.163k (± 6.1%) i/s -      5.850k in   5.049510s
   sequel_pluck_sary      1.291k (± 7.2%) i/s -      6.528k in   5.084559s
   sequel_pluck_tary    233.442  (± 3.0%) i/s -      1.173k in   5.030128s
```

This PR without sequel_pg:
```
   sequel_pluck_iary      3.146k (± 7.5%) i/s -     15.650k in   5.001945s
   sequel_pluck_sary      2.974k (± 9.0%) i/s -     14.994k in   5.077874s
   sequel_pluck_tary    260.419  (± 4.2%) i/s -      1.300k in   5.000711s
```

sequel-5.8.0 with sequel_pg:
```
   sequel_pluck_iary      2.605k (± 7.4%) i/s -     12.954k in   5.001034s
   sequel_pluck_sary      2.827k (± 7.2%) i/s -     14.250k in   5.066844s
   sequel_pluck_tary    261.245  (± 4.6%) i/s -      1.326k in   5.086690s
```

This PR with sequel_pg:
```
   sequel_pluck_iary      3.148k (± 7.4%) i/s -     15.759k in   5.032233s
   sequel_pluck_sary      2.929k (± 7.0%) i/s -     14.637k in   5.022570s
   sequel_pluck_tary    252.855  (± 4.4%) i/s -      1.275k in   5.052964s
```




> Have you run the spec_postgres rake task with this change?

I did it now and get 6 failures related to timestamp offsets (probably because of a different local time zone on my computer) and this one:
```
  7) Failure:
PostgreSQL array handling#test_0004_insert and retrieve arrays of other types [/home/local/COMCARD-NT/kanis/Projekte/sequel/spec/adapters/postgres_spec.rb:2271]:
--- expected
+++ actual
@@ -1 +1 @@
-[{:ba=>[#<Sequel::SQL::Blob:0xXXXXXX bytes=2 content="a\u0000">, nil], :tz=>[#<Sequel::SQLTime 10:20:30>, nil], :o=>[1, 2, 3]}]
+[{:ba=>[#<Sequel::SQL::Blob:0xXXXXXX bytes=2 content="a\x00">, nil], :tz=>[#<Sequel::SQLTime 10:20:30>, nil], :o=>[1, 2, 3]}]
```

It expects a UTF-8 encoded Blob, which is probably a failure in the test?

jeremyevans:
Thanks for the benchmarks and benchmarks code.  I'll take a deeper look at this today.

In the long term, no, I don't plan on migrating sequel_pg into pg, though there are may be cases where Sequel may use pg features either instead of sequel_pg features or when sequel_pg is not available and pg is.

I don't think the failure is related to UTF-8 vs. binary, since `"\u0000".encode('UTF-8') == "\x00".b` is `true`.  That failure is more likely related to `SQLTime` differences that aren't visible from inspect output.

jeremyevans:
Looks like the difference for strings is that `pg` uses a faster method of setting encoding on strings (PG_ENCODING_SET_NOCHECK).  If I make sequel_pg use that, I get exactly the same performance (which I should, since both pg's and sequel_pg's implementations are based on the same algorithm.

The lower performance on iary is because sequel_pg currently uses procs for array element conversions for non string arrays. That's probably fixable by reusing sequel_pg's default scalar typecasting for array elements.

So I think this will lead to some optimizations in sequel_pg.  I'm still considering whether to use pg as a fallback if sequel_pg is not available for the array parsing case.

jeremyevans:
Because of the pg's current behavior regarding malformed arrays, I don't think I can accept this currently.  However, after pg-2.0 is released, I think it will be acceptable to use pg's array parsing if sequel_pg is not available (basically, try sequel_pg, then pg, then the pure ruby version).

I want to thank you for submitting this pull request, as it encouraged me to take another look at sequel_pg performance.  I ended up borrowing a macro and function from pg in order to speed up sequel_pg's conversion of string and integer data types, and modified sequel_pg to transparently handle parsing PostgreSQL arrays for common data types (avoiding the overhead of creating a ruby string for the entire array).  I'll be putting out a sequel_pg release later today with those changes.

PR #1526, opened 2018-06-18
Add :top as an option in the list plugin

celsworth:
This adds an option to the list plugin to allow overriding the default of 1 for the top of the list position.

I'm not sure whether whether this could do with more tests, I'd be happy to add them if you let me know.


jeremyevans:
Thanks for the patch! This looks like a good starting point, I'll probably just make a couple changes after merging.  Should have this committed later today.

PR #1529, opened 2018-07-17
Add `Model.first` method for StaticCache plugin

AlexWayfer:
It supports Integers, Hashes and blocks, but not Sequel literals.


AlexWayfer:
```ruby
@db[:t].first(Sequel.lit('id = ?', 2)).must_equal @c2 #=> failed, actual :id=>1
```

Looks broken not in this PR.

lib/sequel/plugins/static_cache.rb
```diff
@@ -80,6 +81,43 @@ def all
...
+          when 0
+            return res.first
+          when 1
+            arg = args[0]
+            case arg
+            when Integer
+              res[0...arg]
+            when Hash
+              res.find do |instance|
+                arg.all? { |key, value| instance[key] == value }
```
AlexWayfer: Dangerous for `id: '2'`: will not work, now it works.

jeremyevans:
The breakage is because the extension specs are run on a mock database, you are responsible for setting the values to return correctly using the mock database API.

Looking at the code, I think the use with a block is incorrect, and the use with a Hash will not handle all cases correctly.  The use with no argument or with a single integer argument without a block should be safe.  For performance reasons, both of those cases should use `@all`, not `all` so that an extra array is not allocated.

If you want something similar to the Hash support, where a database query is not issued, you can do:

```
@db[:t].to_enum.find{|d| d.id == 2}
```

I'm open to there being a new method added to the plugin that does something like that.

AlexWayfer:
> Looking at the code, I think the use with a block is incorrect

I did it as written in the documentation for [`Dataset#first`](https://sequel.jeremyevans.net/rdoc/classes/Sequel/Dataset.html#method-i-first) (there is no block variable), if you about public interface.

> and the use with a Hash will not handle all cases correctly

All cases from `Dataset#first` documentation, I hope.

> If you want something similar to the Hash support, where a database query is not issued, you can do:

I want to have the similar (common) method for any models, cached or not.

I think it's not too hard to write method for cached models similar to regular models (with some notes).

> For performance reasons, both of those cases should use @all, not all so that an extra array is not allocated.

OK, done.

> The breakage is because the extension specs are run on a mock database, you are responsible for setting the values to return correctly using the mock database API.

OK, I got it, thank you. Fixed.

jeremyevans:
Just because the code works for the small number of examples in the documentation does not mean it handles all cases correctly.  You are using `instance_exec` for the block instead of evaluating the block as a virtual row block.  Your code to handle the hashes is very limited in terms of what it handles correctly.  For example, in Sequel. `first(:id=>1)`  is equivalent to `first(1=>:id)`, and `first(:id=>'1')` is very different than `first(:id=>Sequel.lit('1'))`

You may think it isn't too hard to write a method that works for both cached and regular models.  Some people think they can parse SQL with regular expressions and get something that works. I'm sure both are true for some definitions of "works".

I appreciate that you want the cached behavior to be the same as the non-cached behavior, but you basically cannot use the cache any time you have an SQL expression, and with `first`, both blocks and hashes generate SQL expressions.

AlexWayfer:
> Just because the code works for the small number of examples in the documentation does not mean it handles all cases correctly. 

I understand, but this method can be complemented by another examples and use-cases. I just implemented basic (I think it's better than nothing, but I'm not sure).

> You are using instance_exec for the block instead of evaluating the block as a virtual row block. Your code to handle the hashes is very limited in terms of what it handles correctly. For example, in Sequel. first(:id=>1) is equivalent to first(1=>:id), and first(:id=>'1') is very different than first(:id=>Sequel.lit('1'))

I know, it's very hard (maybe even impossible) to simulate SQL-like behavior. But I'm happy with basic usage, and I think somebody can be happy too.

> You may think it isn't too hard to write a method that works for both cached and regular models.

No, I think it isn't too hard to write a method that works for both cached and regular models _in basic use-cases_. And I think that support for basic use-cases is better than no support at all. But, again, I'm not sure about this.

> I appreciate that you want the cached behavior to be the same as the non-cached behavior, but you basically cannot use the cache any time you have an SQL expression, and with first, both blocks and hashes generate SQL expressions.

Of course you can reject this PR, and then I'll leave around of 10 LoC in our project in conditional. Or you give some corrections and I'll try to implement them.

I just faced with situation when our project has some common-model code, and model can be cached or not. And I thought that somebody else can be faced with something similar. And I'm suggesting this partially-similar-method-for-basic-usage. :man_shrugging:

jeremyevans:
We have a difference of opinion on this topic I guess.  I think we agree it's infeasible to do what you want correctly. I don't want to add support for something that will be broken in many use cases and just happens to work for your use cases.

The static_cache plugin should not change the behavior of `first`.  In the cases where it can ensure the method arguments are handled the same way, with better performance, it can implement optimizations to improve performance.  But if it can't ensure that, it should call super to get the default behavior.

I don't want to reject this PR, as I think the no-argument case and the single Integer argument case without a block can be handled correctly.  I already mentioned that to correct the PR, you need to remove the current handling of blocks and hashes and call super in those cases.

Note that the `to_enum.find` approach will work in both cached and regular mode, though performance in regular mode may leave something to be desired.  If you want to have a method like `first` that works in memory for cached models and produces a suitable query for regular models, that's fine, but you should limit the allowed API, and that's something you'll have to support as an external plugin, it's not something I would want to ship in Sequel.

AlexWayfer:
> I think we agree it's infeasible to do what you want correctly.

Yep.

> I don't want to add support for something that will be broken in many use cases and just happens to work for your use cases.

OK.

> The static_cache plugin should not change the behavior of first. In the cases where it can ensure the method arguments are handled the same way, with better performance, it can implement optimizations to improve performance. But if it can't ensure that, it should call super to get the default behavior.

I got it. But… what about `Model.count`? There is "(without an argument or block)".

> I don't want to reject this PR, as I think the no-argument case and the single Integer argument case without a block can be handled correctly. I already mentioned that to correct the PR, you need to remove the current handling of blocks and hashes and call super in those cases.

OK, done.

> Note that the to_enum.find approach will work in both cached and regular mode, though performance in regular mode may leave something to be desired.

OK, I'll look into it, thank you.

> If you want to have a method like first that works in memory for cached models and produces a suitable query for regular models, that's fine

Yeah, for Hashes.

> but you should limit the allowed API, and that's something you'll have to support as an external plugin, it's not something I would want to ship in Sequel.

OK, thank you.

jeremyevans:
`Model.count` is handled correctly by static_cache I think.  Without an argument or block, it gives the number of cached records.  With an argument or block, you get the default behavior.

The new code looks mostly good, except that if a block is given then super should be returned, and there should be new specs added for `first` instead of modifying the existing specs. The specs should include all `first` usage, including usage that calls super.

AlexWayfer:
Thank you.

> Model.count is handled correctly by static_cache I think. Without an argument or block, it gives the number of cached records. With an argument or block, you get the default behavior.

OK, I understood.

> The new code looks mostly good, except that if a block is given then super should be returned, and there should be new specs added for first

Yep, you're right. Added, code fixed.

Also, I changed documentation for this method. I hope, correctly.

> instead of modifying the existing specs.

Thanks, returned.

> The specs should include all first usage, including usage that calls super.

I did everything what remembered.

----

I have another code for this method, with more complex condition but with one `super` instead of two. Here it's, FYI:

```ruby
return @all.first if args.length.zero? && block.nil?
if args.length > 1 || !(arg = args.first).is_a?(Integer) || block
  return super
end
@all[0...arg]
```

jeremyevans:
Thanks, this looks good, I'll try to merge and test later today.

PR #1533, opened 2018-08-17
Make `Model.load_cache` method public in `StaticCache` plugin

AlexWayfer:
Now we can reload the cache easy!


jeremyevans:
Thanks for the patch. I think this makes sense.  `load_cache` is already called by the `pg_static_cache_updater` extension.

AlexWayfer:
>  load_cache is already called by the pg_static_cache_updater extension.

Oh, OK, I've replaced `send(:load_cache)` with a direct call.

PR #1534, opened 2018-08-22
Add clob to mapping for bind params oracle

pipistrellka:
I can't execute oracle procedure with CLOB value


jeremyevans:
Thanks for the patch. This code change looks fine to me. Could you add a spec to `spec/adapters/oracle.rb` to test it?

pipistrellka:
I added soon, thank you.

jeremyevans:
I added a spec for this in 66541f1c9a96def6c3da87d3bf412d85b981752b, which passes without this patch.  I'm not sure in which cases this patch is necessary.  I'm going to close this now, if you believe this is still necessary, please post an update with spec changes that fail without the patch.

jeremyevans:
Thanks, I'm going to reopen this and see if I can get the spec working on the jdbc adapter as well.

PR #1536, opened 2018-09-04
Fix Rdoc indent

janko-m:
This removes an extra indent that was probably unintentional.

screen shot 2018-09-04 at 04 19 29


jeremyevans:
Good catch, thanks!

PR #1539, opened 2018-09-10
Fix warning: shadowing outer local variable - i

vais:


jeremyevans:
Thanks for the patch!  This looks good, I'll test and merge tomorrow.

PR #1540, opened 2018-09-12
RDoc fixes

janko-m:
The first commit fixes a typo in RDocs. Second commit makes some references to the `#column` method links that you can follow, which I found I'd like to have.


jeremyevans:
Looks good, thanks!

PR #1542, opened 2018-09-13
Column names in mass assignment docs fixed

vlado:


jeremyevans:

PR #1543, opened 2018-09-14
Allow `Model.all` to receive a block, as `Dataset#all`

AlexWayfer:


jeremyevans:
Thanks for working on this.  There seem to be a few changes here that are not all related, let me go over each change.

1) all accepts block:  This slows down the common case where the cache is frozen, as map is slower than dup.  Note that .to_a is not necessary after map, as map returns an array.  I'm not opposed to the idea, but maybe just add an `elsif block; map(&block)` to the existing code so as not to lose the dup optimization.

2) map refactoring: this eliminates one of the branches but does not change the behavior, so it should be safe.  However, please revert the formatting change as it makes the code less consistent.

3) load_cache not freezing objects in all cases: cached objects should always be frozen.  If frozen: false is used, you can update uncached objects, but you should never be able to update cached objects.  This part should be removed.

AlexWayfer:
> This slows down the common case where the cache is frozen, as map is slower than dup.

Yep, confirmed.

> Note that .to_a is not necessary after map, as map returns an array.

No, it can return `Enumerator` without block at all, as I know.

> I'm not opposed to the idea, but maybe just add an elsif block; map(&block) to the existing code so as not to lose the dup optimization.

Done.

> map refactoring: this eliminates one of the branches but does not change the behavior

It uses `@all` instead of `self.all`.

> However, please revert the formatting change as it makes the code less consistent.

What formatting are you talking about?

-----

Oh, I got it… Done.

Almost everything reverted except `block` for `all`. Current code is good enough.

jeremyevans:
Thanks for the quick changes. This looks good, I'll test and merge tomorrow.

PR #1545, opened 2018-09-15
Add examples for #association_join, #eager and #eager_graph

janko-m:
Since it's not clear from reading the RDocs of `#association_join` what kind of arguments it receives (the same kinds as `#eager` and `#eager_graph`), I thought it would be good to add examples to RDocs.

While here, I also often missed examples in `#eager` and `#eager_graph` RDocs, so I added them as well. I know these examples are already present when you scroll to the top of that RDoc document, but many people might know that, and I think it's good to have a bit of redundancy in the documentation.

### DIFF
--- a/lib/sequel/model/associations.rb
+++ b/lib/sequel/model/associations.rb
@@ -2641,7 +2641,7 @@ module Sequel
       #   Album.eager(:artist, :genre).all
       #   Album.eager_graph(:artist, :genre).all
       #   Album.eager(:artist).eager(:genre).all
-      #   Album.eager_graph(:artist).eager(:genre).all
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for working on this!  I agree that some redundancy is OK in the documentation.

AlexWayfer:
Please note, that these methods also can receive `proc` as Hash value:

```ruby
Album.association_join(artist: ->(ds) { ds.where(age > 20) }).sql
# SELECT * FROM "albums" INNER JOIN (SELECT * FROM "artists" WHERE ("age" > 20)) ON ("albums"."artist_id" = "artists"."id")
```

PR #1546, opened 2018-09-16
Show association filtering for #association_join, #eager and eager_graph

janko-m:
Addressing https://github.com/jeremyevans/sequel/pull/1545#issuecomment-421734619. I also added some short descriptions as I thought it would be useful, especially for clarifying the filter blocks. This also removes some superfluous `.all` calls in the `#association_join` examples.


jeremyevans:
Thanks, I agree it makes sense to mention this feature in the method RDoc.

PR #1547, opened 2018-09-17
Support single_value type in prepared statements

rintaun:
This pull request adds `:single_value` as a supported type for prepared statements, allowing single-column queries for counts, averages, etc. to be prepared.


jeremyevans:
Thanks for the patch!  This looks good, I'll test and merge tomorrow.

PR #1549, opened 2018-09-30
Feature: Enable dataset condition inversion using NOT operator

rintaun:
Currently when a dataset is inverted via `.invert`, it inverts each individual operator within the condition.  This allows inversion to be completed by wrapping the entire expression in a NOT clause.

I imagine there is more that needs to be done with this prior to merging. I will be adding test cases, but for the moment, is there anything else this needs to do?

### DIFF
--- a/lib/sequel/dataset/query.rb
+++ b/lib/sequel/dataset/query.rb
@@ -456,15 +456,20 @@ module Sequel
     #
     # See documentation for exclude for how inversion is handled in regards
     # to SQL 3-valued boolean logic.
-    def invert
+    def invert(opts=OPTS)
       cached_dataset(:_invert_ds) do
         having, where = @opts.values_at(:having, :where)
         if having.nil? && where.nil?
           where(false)
         else
           o = {}
-          o[:having] = SQL::BooleanExpression.invert(having) if having
-          o[:where] = SQL::BooleanExpression.invert(where) if where
+          if opts[:not]
+            o[:having] = SQL::BooleanExpression.not(having) if having
+            o[:where] = SQL::BooleanExpression.not(where) if where
+          else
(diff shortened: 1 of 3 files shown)


AlexWayfer:
> I will be adding test cases, but for the moment, is there anything else this needs to do?

Docs, maybe?

jeremyevans:
Thanks for the patch, but I think this feature is best changed at a different level.  It needs to be a global extension that overrides `Sequel::SQL::BooleanExpression.invert` to wrap the expression in `NOT` instead of pushing down the inversion.

It would have to have specs run in a separate process similar to the core extension specs.  There should also be an environment variable for the integration/adapter tests that enables the extension, so tests can be run to make sure it doesn't break things.

jeremyevans:
Actually, since the underlying issue is limited to ANY/SOME/ALL, we may be able to get away with:

``` diff
diff --git a/lib/sequel/sql.rb b/lib/sequel/sql.rb
index 45904ef87..236453383 100644
--- a/lib/sequel/sql.rb
+++ b/lib/sequel/sql.rb
@@ -1093,6 +1093,8 @@ module Sequel
           Sequel.delay{|ds| from_value_pair(l, r.call(ds))}
         when Dataset::PlaceholderLiteralizer::Argument
           r.transform{|v| from_value_pair(l, v)}
+        when Function
+          Wrapper.new(new(:'=', l, r))
         else
           new(:'=', l, r)
         end
```

That won't make tomorrow's release, but it doesn't appear to break any specs, so I should be able to merge it after the release tomorrow after adding specs for inverting equality expressions with function arguments.

rintaun:
That is certainly a much more elegant solution!

PR #1550, opened 2018-10-01
Add `Sequel::Model#match?` alias for `#===`

AlexWayfer:
For example, Ruby core has `Regexp#===` and `Regexp#match?`.

[Ruby Style Guide](https://github.com/rubocop-hq/ruby-style-guide#no-case-equality) says to avoid `===`,
and such cop is enabled in RuboCop by default. But it's hard to recognize Sequel cases,
so `Sequel::Model#===` can be just replaced with `#match?`.

Also fix example for `#===`.


jeremyevans:
I don't think `match?` is a good method name for this.  Possibly a better name would be `primary_key_equal?`, but I don't like adding aliases unless the current method name is misleading.  I don't think that is true in the case, but I'll agree that `===` is not descriptive as to what the method does.

Can you please make the following changes:

* Use `primary_key_equal?` instead of `match?`.
* Not use `shared_description` in the spec, since I think normal module inclusion will work (shared_description is only needed in cases where normal module inclusion doesn't use the desired execution order for before/after hooks).

AlexWayfer:
> Possibly a better name would be `primary_key_equal?`

On the one hand, I like detailed names. On the other, this name seems too long.

What about `pk_equal?` ?

I looked at existing methods, and found:
* [`Sequel::Model::InstanceMethods`](https://sequel.jeremyevans.net/rdoc/classes/Sequel/Model/InstanceMethods.html) has `#pk`, `#pk_hash` and `#qualified_pk_hash` methods;
* but [`Sequel::Model::ClassMethods`](https://sequel.jeremyevans.net/rdoc/classes/Sequel/Model/ClassMethods.html) has `.primary_key`, `.primary_key_hash` and `.qualified_primary_key_hash` methods.

I like consistency, and I'm not sure, which name is better, and do we should somehow lead instance and class methods to a common naming.

jeremyevans:
Since this is an instance method, `pk_equal?` is probably better.

AlexWayfer:
> Not use shared_description in the spec, since I think normal module inclusion will work (shared_description is only needed in cases where normal module inclusion doesn't use the desired execution order for before/after hooks).

I'm sorry, but how to do this? Do you have some examples? Because `::Model` has no `it` method, and I didn't find examples in existing code.

> Since this is an instance method, `pk_equal?` is probably better.

Also, `pk_eql?` can be even better.

`::Object` has `eql?`, `==` and `equal?` methods.

And `equal?` is not recommended for overriding: 

> Unlike ==, the equal? method should never be overridden by subclasses as it is used to determine object identity (that is, a.equal?(b) if and only if a is the same object as b)

http://ruby-doc.org/core-2.5.1/Object.html#method-i-eql-3F

But primary keys are compared via `==`, and `eql?` can be alias for this:

> For objects of class `Object`, `eql?` is synonymous with `==`. Subclasses normally continue this tradition by aliasing `eql?` to their overridden `==` method, but there are exceptions.

What do you think?

AlexWayfer:
> I think pk_equal? is fine, and pk_eql? is too terse.

OK, `#match?` replaced with `#pk_equal?`. Commit message, PR title and description changed.

> it's probably simplest to just use an array with each

Thank you, done.

jeremyevans:
Thanks, looks good, will test and merge shortly.

PR #1554, opened 2018-10-04
Make roots and roots_dataset to dataset methods in tree plugin

JelF:
It is usefull when used with associations.
I.e. `foo.bars_dataset.roots`


jeremyevans:
Thanks for the patch.  I'll make a few tweaks to this, but I should be able to merge it today.

PR #1555, opened 2018-10-05
Upsert: Explain how to upsert using existing column value

EdgarOrtegaRamirez:
Hello @jeremyevans, faced this issue today. Hope it can help others


jeremyevans:
Thanks for the patch. I agree that an additional example would be helpful here.

PR #1557, opened 2018-10-08
Improve Sequel.case documentation

janko-m:
I've often forget the `Sequel.case` arguments, and when I would look at the docs it would take me some time to figure out which argument is for what. So I updated the documentation to explain each argument, and to go from simpler to more complex examples.

I also added the hash variant for the sub-expression example, as I find it much more intuitive than an array. It also makes it clearer that it's just an extension of the "simple column value" example.


jeremyevans:
Thanks for the patch!

PR #1559, opened 2018-10-18
Normalize some RDocs

janko-m:
There are some examples which assume that core extensions are loaded, so I updated those to use `Sequel[]`. I also replaced some `Sequel.lit` instances with `Sequel[]` for consistency with other examples.


jeremyevans:
Thanks for the patch, we definitely want to remove examples using the core extensions.

PR #1561, opened 2018-10-18
Add support for qualifying hashes in filters

janko-m:
If we had a joined dataset where we wanted to filter by many columns on the other table, currently we could explicitly qualify each column using `Sequel[]`:

```rb
Track
  .association_join(:album)
  .where(
    Sequel[:albums][:released_at] => Date.today,
    Sequel[:albums][:name]        => "Back to Black",
    Sequel[:albums][:single]      => true,
  )
```

or virtual rows:

```rb
Track
  .association_join(:album)
  .where{
    (albums[:released_at] =~ Date.today)      &
    (albums[:name]        =~ "Back to Black") &
    (albums[:single]      =~ true)
  }
```

Since that can get a bit repetitive if we have a lot of column filters, and we might want to keep using hashes, we can rewrite it using `Sequel.deep_qualify`:

```rb
Track
  .association_join(:album)
  .where(Sequel.deep_qualify(:albums,
    released_at: Date.today,
    name:        "Back to Black",
    single:      true,
  ))
```

This pull request adds a syntax sugar for the latter `Sequel.deep_qualify` version, allowing us to use a nested hash instead, where the key is the table name and the value is the expression we want to deep qualify (like in Active Record):

```rb
Track
  .association_join(:album)
  .where(albums: {
    released_at: Date.today,
    name:        "Back to Black",
    single:      true,
  })
```

I think it would be a very convenient syntax to have, I found myself wishing to use it. Let me know if you're open to accepting this feature. In that case I'll go ahead and update the documentation.

### DIFF
--- a/lib/sequel/sql.rb
+++ b/lib/sequel/sql.rb
@@ -1088,6 +1088,9 @@ module Sequel
         when ::Array
           r = r.dup.freeze unless r.frozen?
           new(:IN, l, r)
+        when ::Hash
+          r = r.dup.freeze unless r.frozen?
+          Sequel.deep_qualify(l, from_value_pairs(r))
         when ::String
           r = r.dup.freeze unless r.frozen?
           new(:'=', l, r)
--- a/spec/core/expression_filters_spec.rb
+++ b/spec/core/expression_filters_spec.rb
@@ -62,6 +62,14 @@ describe "Blockless Ruby Filters" do
     @d.l(expr).must_equal '(b IN (1))'
   end
 
+  it "should not modifying boolean expression created from hash if hash is modified" do
+    a = {:c=>'d'}
+    expr = Sequel.expr(:b=>a)
+    @d.l(expr).must_equal "(b.c = 'd')"
+    a[:e] = 'f'
+    @d.l(expr).must_equal "(b.c = 'd')"
+  end
+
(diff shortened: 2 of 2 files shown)


jeremyevans:
Sorry, but this can't be supported, as the syntax already means something else (assume column `a` is boolean and `b` is integer):

```ruby
DB[:table].where(:a=>{:b=>1})
# SELECT * FROM table WHERE (a = (b = 1))
```

This is something that could be supported via an extension using a different method.  Example:

```ruby
Track
  .association_join(:album)
  .qualified_where(album: {
    released_at: Date.today,
    name:        "Back to Black",
    single:      true,
  })
```

I'm not sure whether I'd like to ship such an extension with Sequel or not.  I'd like to get feedback from the community before making a decision, so please bring it up on the Google Group if you would like to see it added.

janko-m:
Thanks for the suggestion.

Yeah, I knew that this syntax already translates to something, but I always assumed it wasn't valid SQL, or at least that it was very unusual. I'll ask on the google group.

AlexWayfer:
> I'll ask on the google group.

I didn't find, so I've created: https://groups.google.com/forum/#!topic/sequel-talk/qhBAh7r0AKY

PR #1565, opened 2018-11-08
Fix default class table inheritence alias when dataset is qualified

benalavi:
From discussion here: https://groups.google.com/forum/#!topic/sequel-talk/Awt--16UGLw

I used your suggestion of setting `@cti_alias` as the unqualified identifier rather than coercing into a string, I think that made more sense than my naive implementation.


jeremyevans:
Thanks for the patch.  This looks good, I'll test and merge tomorrow.

PR #1566, opened 2018-11-09
Allow optional default value in Sequel.case

janko-m:
The ELSE clause in a CASE statement is optional, but in Sequel it's currently required. This allows omitting the default value.

### DIFF
--- a/lib/sequel/dataset/sql.rb
+++ b/lib/sequel/dataset/sql.rb
@@ -307,8 +307,10 @@ module Sequel
         sql << t
         literal_append(sql, r)
       end
-      sql << " ELSE "
-      literal_append(sql, ce.default)
+      if ce.default?
+        sql << " ELSE "
+        literal_append(sql, ce.default)
+      end
       sql << " END)"
     end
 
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch.

When I originally added this code, I made a deliberate decision to require an ELSE value, even those the ELSE clause is optional in SQL.  There are few cases where you wouldn't want to use ELSE, and in those cases you can just use `nil` as the default value (as SQL returns NULL in that case).  I felt it was more likely people would accidently forget the ELSE value if the argument was not required, so in terms of a tradeoff, I chose to avoid accidentally misuse with the cost of increased verbosity.

I'm fine revisiting that decision, so if you strongly feel that the ELSE value should not be required, please post on sequel-talk and see what the community thinks about the idea. In terms of the patch itself, you don't need the extra `no_default` magic here, a simple `nil` default value for the second argument would work.  It would use `ELSE NULL` instead of omitting  the ELSE clause, but that should not cause issues in practice as far as I know (and doesn't require branching to handle the no default case).

janko-m:
Thanks for the explanation. After hearing it, and realizing that I can reformat

```rb
Sequel.case(
  {
    foo: "bar",
    baz: "quux",
  },
  nil
)
```

as

```rb
Sequel.case({
  foo: "bar",
  baz: "quux"
}, nil)
```

I don't feel strongly enough about having the ability to omit the default value. You're right that because it's a required argument it makes you think whether you should put the default value.

PR #1571, opened 2018-11-21
Add :qualify_tables option to CTI plugin to automatically qualify sub…

benalavi:
…class tables with same qualifier as superclass when superclass table is qualified


jeremyevans:
Thanks for the patch!  This looks good, I will test and merge next Monday.

PR #1581, opened 2018-12-11
Fix for qualify_tables option to CTI plugin, properly qualifies subcl…

benalavi:
…asses of subclasses


jeremyevans:
Thanks for the patch, will merge and test shortly.

PR #1585, opened 2018-12-24
Multiple migration directories

Lavode:
This PR allows the two migrator classes to look for migration files in multiple directories.

This might be used for either organizational purposes, where - if a project has many migrations - the user might want to group them. Alternatively (as is the case for me) it can be useful if there is a common framework-like core, which is used by many separate applications.

All previous restrictions - ie migration names needing to be unique, and no missing migrations being allowed unless specifically configured - still apply.

I have attempted to guarantee full backwards-compatability - both for users of `bin/sequel` as well as for those using the `IntegerMigrator` / `TimestampMigrator` / `Migrator` classes directly. According to my manual testing this also works - and tests still pass. :)

During development I have also encountered two bugs related to specifying non-existent or empty migration directories, and a few minor typos in test descriptions. These have been fixed in separate commits (`855f98b` and `a4e8467` respectively), should you chose to merge them, but not the main change.

### DIFF
--- a/bin/sequel
+++ b/bin/sequel
@@ -9,7 +9,7 @@ dump_migration = nil
 dump_schema = nil
 dump_indexes = nil
 env = nil
-migrate_dir = nil
+migrate_dirs = []
 migrate_ver = nil
 backtrace = nil
 show_version = false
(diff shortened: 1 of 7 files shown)


jeremyevans:
Thanks for working on this and submitting a patch.  However, I don't think this is the route I would like to take.  In my mind, multiple migration directories should contain independent sets of migrations.  The Migrator API already supports this using a separate migration table per directory (`bin/sequel` does not expose this aspect of the API).

Your use of multiple migration directories could be handled by the current code by creating a temp directory and adding symlinks to each migration file to the other directories, and using the temp directory as the migration directory.

Thanks for including the bug and typo fixes as separate patches, I'm currently testing those changes and should push them shortly.

PR #1594, opened 2019-01-25
Get count of rows affected without issuing a separate query (ADO/MSSQL)

vais:


jeremyevans:
Thanks for working on this.  I'll run it through my ADO/MSSQL test suite and if nothing breaks I'll merge it.  I will probably be able to do that on Monday.

PR #1597, opened 2019-02-08
add sslcert/sslkey options

mistaflible:
I was unable to use Sequel to connect to a postgres instance on Google Cloud SQL until these options were added.

### DIFF
--- a/lib/sequel/adapters/postgres.rb
+++ b/lib/sequel/adapters/postgres.rb
@@ -200,7 +200,9 @@ module Sequel
             :password => opts[:password],
             :connect_timeout => opts[:connect_timeout] || 20,
             :sslmode => opts[:sslmode],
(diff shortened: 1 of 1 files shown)


jeremyevans:
You can use the :driver_options option to set these.  I don't want to add any additional options to the adapter if the same functionality can be handled via :driver_options.  Have you tried using :driver_options?

mistaflible:
Oops, didn't even see that, even though it was two lines down in the code. :driver_options worked just fine. Thanks!

gkirill:
@mistaflible I think this is not about the amount of code, this is more about not having same logic in two different places. You, the DRY principle.

mistaflible:
@gkirill i wasn't commenting about the number of lines of code in the change, I was commenting on how close to my proposed patch the code for the proper solution was. I just didn't see it before submitting my patch.

PR #1601, opened 2019-03-14
describe connection validation query

beanieboi:
Hey,

i saw a crazy amount of `SELECT NULL` in our database activity log and it took me a while to figure out where it came from. This will change the query to include a SQL comment stating where it's coming from. This also makes it easy to take the message and grep the Sequel code to find it's location.

This would have saved me some time.

Let me know what you think :)

### DIFF
--- a/lib/sequel/database/connecting.rb
+++ b/lib/sequel/database/connecting.rb
@@ -326,7 +326,7 @@ module Sequel
 
     # The SQL query to issue to check if a connection is valid.
     def valid_connection_sql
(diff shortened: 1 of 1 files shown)


jeremyevans:
I don't think we want to do this by default.  You should probably just override the method for your Database instance:

```ruby
def DB.valid_connection_sql
  "SELECT NULL /* From Ruby Sequel */"
end
```

BTW, it looks like you intended to use a comment, but you are actually selecting a string with the text of the comment.

PR #1602, opened 2019-03-17
Remove JRuby check from excluded dedup spec.

headius:
This spec passes fine on JRuby since jruby/jruby#5190.


jeremyevans:
Thanks for the patch.  This would make the spec still fail on JRuby 9.2.0.0, but at this point I don't think we should worry about that.

jeremyevans:
Cherry-picked as 8fe03b3020dfd402e3c4977b776078d6d5305626.

PR #1603, opened 2019-03-18
Add `#rename_enum_value` method for `pg_enum` extension

AlexWayfer:
This ability was added into PostgreSQL 10: https://commitfest.postgresql.org/10/588/

### DIFF
--- a/lib/sequel/extensions/pg_enum.rb
+++ b/lib/sequel/extensions/pg_enum.rb
@@ -17,6 +17,12 @@
 #
 #   DB.rename_enum(:enum_type_name, :enum_type_another_name)
 #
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch.  This looks good to me.  Can you also update the pg_enum spec in the postgres adapter spec file to include testing this?  That way this gets tested against a real PostgreSQL database.

AlexWayfer:
> Can you also update the pg_enum spec in the postgres adapter spec file to include testing this? That way this gets tested against a real PostgreSQL database.

Yes, of course! Done.

jeremyevans:
Thanks, I will merge and test shortly.

PR #1607, opened 2019-03-29
Document the :connect_sqls option

vais:
This option is easier to use than `:after_connect` if all you need is to run some SQL on each new connection (e.g. when `SET`-ting sane defaults on a tiny_tds connection to MS SQL Server). Not sure why this option is undocumented - feel free to ignore if that's on purpose... Thanks!


jeremyevans:
This documentation says at the bottom: "All options given are also passed to the connection pool.".  The ConnectionPool documents the :connect_sqls option (as well as the :after_connect option).  For some reason, both document the :preconnect option.  The connection pool probably shouldn't document preconnect, as it does not handle it, the database code does.  I'll see if I can fix that.

vais:
👍

PR #1609, opened 2019-04-11
Added support for 'float unsigned' in the schema_dumper extension.

keeguon:
I was playing around with the "schema_dumper" earlier today and I noticed that some of my "float unsigned" column (coming from MySQL) were wrongly converted into strings because it didn't match the regexp for the Float type so fixed it.


jeremyevans:
Thanks for the patch.  This looks good, but can you please update the spec as well?

keeguon:
@jeremyevans here you go !

keeguon:
@jeremyevans out of curiosity, when do you plan to push a new release on rubygems with the above included ?

jeremyevans:
If you look at the release history (https://rubygems.org/gems/sequel/versions), you can probably guess the approximate date. :)

PR #1622, opened 2019-05-30
Adds sequel-through plugin

kenaniah:
This is a plugin I wrote to support an easy syntax for creating model through associations.


jeremyevans:
Looks good, thanks!

PR #1630, opened 2019-07-04
Improve dataset filtering documentation when mixing AND and OR

francois:
I was looking for documentation on how to combine AND and OR filtering in the same dataset, but couldn't find anything on 

* http://sequel.jeremyevans.net/rdoc/files/doc/querying_rdoc.html
* http://sequel.jeremyevans.net/rdoc/files/doc/dataset_filtering_rdoc.html

I did find a recent discussion on the mailing list:

https://groups.google.com/forum/#!searchin/sequel-talk/where$20or%7Csort:date/sequel-talk/sNEUId8J9Vw/1yJteFuwAwAJ

This PR adds the documentation where I would eventually have found it. Hope this helps!

---
This change is [Reviewable](https://reviewable.io/reviews/jeremyevans/sequel/1630)

### DIFF
--- a/doc/dataset_filtering.rdoc
+++ b/doc/dataset_filtering.rdoc
@@ -106,6 +106,21 @@ You can also use placeholders with :placeholder and a hash of placeholder values
   items.where(Sequel.lit('category = :category', category: "ruby")).sql
   # "SELECT * FROM items WHERE category = 'ruby'"
 
+In order to combine AND and OR together, you have a few options:
+
+  items.where(category: nil).or(category: "ruby")
+  # SELECT * FROM items WHERE (category IS NULL) OR (category = 'ruby')
+
+This won't work if you add other conditions:
+
+  items.where(name: "Programming in Ruby").where(category: nil).or(category: 'ruby')
+  # SELECT * FROM items WHERE ((name = 'Programming in Ruby') AND (category IS NULL)) OR (category = 'ruby')
+
+The OR applies globally and not locally. To fix this, use & and |:
+
+  items.where(Sequel[name: "Programming in Ruby"] & (Sequel[category: nil] | Sequel[category: "ruby"]))
+  # SELECT * FROM items WHERE ((name = 'Programming in Ruby') AND ((category IS NULL) OR (category = 'ruby')))
+
 === Specifying SQL functions
 
 Sequel also allows you to specify functions by using the Sequel.function method:


doc/dataset_filtering.rdoc
```diff
@@ -106,6 +106,21 @@ You can also use placeholders with :placeholder and a hash of placeholder values
...
+  # SELECT * FROM items WHERE (category IS NULL) OR (category = 'ruby')
+
+This won't work if you add other conditions:
+
+  items.where(name: "Programming in Ruby").where(category: nil).or(category: 'ruby')
+  # SELECT * FROM items WHERE ((name = 'Programming in Ruby') AND (category IS NULL)) OR (category = 'ruby')
+
+The OR applies globally and not locally. To fix this, use Sequel.|:
+
+  items.where(name: "Programming in Ruby").where(Sequel.|([:category, nil], [:category, "ruby"]))
```
tycooon: I think the most readable way for this kind of stuff is using sequel epressions, like:
```ruby
where((Sequel[:category] =~ nil) | Sequel[:category] =~ ruby))
```
This can also be achieved with virtual row blocks.

AlexWayfer: Or even:

```ruby
.where(Sequel[category: nil] | Sequel[category: ruby])
```

tycooon: @AlexWayfer  Wow, that's cool stuff, didn't see that before 👍

AlexWayfer: It doesn't work for non-equal operators, like `Sequel[:count] > 0`, but it's very handful, yeah.

Also, you can combine them all, like:

```ruby
.where(Sequel[name: "Programming in Ruby"] & (Sequel[category: nil] | Sequel[category: ruby]))
```

francois:
Do we need to require an extension for `&` and `|` to work?

I much prefer what @AlexWayfer proposes. I'll rework the PR to include that instead.

francois:
Now that I'm reading the documentation in more details, I saw a reference to `Sequel.|` a few lines above, but it never clicked for me that this was the solution to my problem. I still feel like the new text improves documentation.

AlexWayfer:
> Do we need to require an extension for `&` and `|` to work?

No. 

And yeah, I found the docs about `AND` and `OR`: https://sequel.jeremyevans.net/rdoc/files/doc/sql_rdoc.html#label-Boolean+Operators+-28AND+OR-29

`sql` and `querying` are most helpful in queries building for me. I'm even would like to merge them.

jeremyevans:
This looks good, thanks for the patch!

PR #1633, opened 2019-07-16
More readability for Change Log using rdoc extension

anup1710:
Renamed the CHANGELOG and CHANGELOG.old files to use rdoc extension and updated the reference.


jeremyevans:
I can understand that rdoc formatting makes it look a little nicer. You may not be aware of it, but the CHANGELOG is already formatted using rdoc on the website (http://sequel.jeremyevans.net/rdoc/files/CHANGELOG.html).  I don't think the benefits of changing the file extension in the repository outweigh the costs, so I'm going to close this.

AlexWayfer:
> outweigh the costs

Can you please call the costs? PR is ready, something else? I just want to understand.

jeremyevans:
Every commit has a cost.  Every commit makes the repository larger, makes git commands for processing history take longer.  A better question is: What's the benefit?  The file is already formatted using RDoc on the website.  The only advantage I can see is the file is formatted nicer on GitHub, and that isn't important to me.

AlexWayfer:
> Every commit has a cost. Every commit makes the repository larger, makes git commands for processing history take longer. A better question is: What's the benefit? The file is already formatted using RDoc on the website. The only advantage I can see is the file is formatted nicer on GitHub, and that isn't important to me.

OK. I guess, you can't calculate the costs of this commit (additional microseconds of git commands), or it's hard enough (doesn't worth it), and you can't measure GitHub (and/or other platform, like text editors) formatting enhancement  (there are no units for this), so you can't compare and just deciding to decline. OK.

PR #1634, opened 2019-07-17
Fix typo at pg_row doc

Mehonoshin:


jeremyevans:
Thanks for the patch!

PR #1638, opened 2019-08-15
Remove unnecessary localhost fallback

sapio-bdeamer:
Mysql2 already falls back to localhost, and providing a default here actually overrides a possible hostname specified in a defaults_file.


jeremyevans:
Thanks for the patch. Looks like Mysql2 doesn't actually handle this, but a `NULL` host in `mysql_real_connect` (called by Mysql2) assumes localhost, so this should be fine.

Interestingly, it was the author of Mysql2 that added this line (in the initial commit for the mysql2 adapter).

PR #1641, opened 2019-09-05
Add `plural_association_methods` plugin with `add_*s` and `remove_*s` methods

AlexWayfer:
Helpers for mass adding.

Accept only Arrays for additional arguments and composite keys support.

### DIFF
--- a/lib/sequel/plugins/plural_association_methods.rb
+++ b/lib/sequel/plugins/plural_association_methods.rb
@@ -1,0 +1,54 @@
+# frozen-string-literal: true
+
+module Sequel
+  module Plugins
+    module PluralAssociationMethods
+      # The `plural_association_methods` plugin adds `Sequel::Model#add_*s` and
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch. I don't want to add additional modification methods for every *_to_many association by default.  It bloats the number of methods.  I was strongly considering removing mutation methods by default and only adding them if explicitly specified in Sequel 5, but I finally decided it wasn't worth the backwards compatibility breakage.

While I don't want to do this by default, this is something that could be added by an plugin easily.  Can you refactor this to be a plugin?

AlexWayfer:
> While I don't want to do this by default, this is something that could be added by an plugin easily. Can you refactor this to be a plugin?

I thought about this. OK, I can try.

AlexWayfer:
Done, I guess.

jeremyevans:
Thanks for this.  I will try to review and merge tomorrow.

PR #1642, opened 2019-09-06
Add `Sequel::Dataset#any?` and `#include?` methods

AlexWayfer:


jeremyevans:
Sorry, but this would break backwards compatibility, as `any?` and `include?` are defined in Enumerable and Sequel::Dataset includes Enumerable.  These could potentially be added in a dataset extensions.  I actually have a dataset extension for `any?` on my todo list.  I don't plan on adding an extension for `include?`, though.

Do you want to submit `any?` as a dataset extension?  You should probably call `super` instead of using `!empty?` if a block is passed.

AlexWayfer:
> Sorry, but this would break backwards compatibility, as any? and include? are defined in Enumerable and Sequel::Dataset includes Enumerable.

It's sad, but OK.

> I don't plan on adding an extension for `include?`, though.

Why?

> Do you want to submit `any?` as a dataset extension?

To be honest, I've added `any?` mostly for `include?`, so without it — not now. And yeah, I remember that (current) `Sequel::Dataset` has one of `any?` and `empty?` methods, which are both existing in `Enumerable`, but one of them is redefined and makes pretty query. But sometimes I don't remember which one exactly.

I've added `include?` for handy code with complex associations like `artist.published_albums_dataset.include?(album)`, but it's hard to pass just a `Sequel::Model` instance, so more realistic example is `artist.published_albums_dataset.include?(id: album.id)`.

jeremyevans:
> > I don't plan on adding an extension for `include?`, though.
> 
> Why?

I think of the include? argument as a complete member of the set of returned values, not as a filter condition.

Note that these (both `any?` and `include?`) could certainly be in an external extension (Dataset#extension doesn't differentiate between internal and external extensions).  Then you could just release that as a gem and use it.

AlexWayfer:
OK, thank you.

PR #1643, opened 2019-09-11
Add setter for `association_multi_add_remove` plugin

AlexWayfer:
Rename it to `association_multi_add_remove_set`.


jeremyevans:
I think the changes look OK.  I wouldn't rename the plugin though. It does add a setter method, but it's not really a multi_set, and is implemented via multi_add and multi_remove.  Is it possible for you to add the setter method, documentation, and tests without renaming the plugin?

AlexWayfer:
> but it's not really a multi_set

For me there are parallels, but OK.

> Is it possible for you to add the setter method, documentation, and tests without renaming the plugin?

Yes, it's possible. I'll do it in a short time.

AlexWayfer:
> Is it possible for you to add the setter method, documentation, and tests without renaming the plugin?

Done.

PR #1657, opened 2019-10-29
New ForeignKeyConstraintViolation errno in MySQL 8

rianmcguire:
MySQL 8.0.13 returns different error numbers for foreign key constraint errors under some conditions:
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-13.html#mysqld-8-0-13-errors

Sequel should recognise these as `ForeignKeyConstraintViolation` errors too.

### DIFF
--- a/lib/sequel/adapters/utils/mysql_mysql2.rb
+++ b/lib/sequel/adapters/utils/mysql_mysql2.rb
@@ -55,7 +55,7 @@ module Sequel
             NotNullConstraintViolation
           when 1062
             UniqueConstraintViolation
-          when 1451, 1452
+          when 1451, 1452, 1216, 1217
             ForeignKeyConstraintViolation
           when 4025
             CheckConstraintViolation


jeremyevans:
Thanks for the patch.  This looks good, I'll try to test and merge tomorrow.

jeremyevans:
Are you sure that `1217` is related to foreign key constraint violations? `1216` definitely appears to be related. I'm seeing the following for `1217`, all of which do not appear to be related to foreign key constraint violations:

https://github.com/mysql/mysql-server/blob/91a17cedb1ee880fe7915fb14cfd74c04e8d6588/storage/ndb/test/ndbapi/testNdbApi.cpp#L290

https://github.com/mysql/mysql-server/blob/91a17cedb1ee880fe7915fb14cfd74c04e8d6588/storage/ndb/src/kernel/blocks/dblqh/Dblqh.hpp#L333

https://github.com/mysql/mysql-server/blob/91a17cedb1ee880fe7915fb14cfd74c04e8d6588/storage/ndb/src/ndbapi/ndberror.cpp#L291

jeremyevans:
Sorry, I should probably have read more closely the documentation you linked to, which does describe both error codes.  Will merge shortly.

PR #1659, opened 2019-10-30
Remove some pound symbols in sql_comments docs

janko:


jeremyevans:
Thanks for the patch!

PR #1664, opened 2019-12-19
Add: IRC Link

BKSpurgeon:
## Why this PR? 
Makes it easier for readers. Readers may ask: Which IRC? Where? How do I get on it? Hopefully the link will take care of that.


AlexWayfer:
You've edited 28th line of file. Look at the 25th line. Also, it'd be better to add all mentioned links if you want to do so (I'm about Google Group).

jeremyevans:
Considering that the IRC info is included a few lines above, I don't think this change is worth it.

BKSpurgeon:
I see. I had actually missed the links just above it.

PR #1667, opened 2019-12-30
Add: DB constant definition

BKSpurgeon:
### Why this PR?
This may help out some inexperienced users migrating to Sequel from ActiveRecord. Otherwise they may spend a fair amount of time trying to make it work (as I did). I hope it helps.


jeremyevans:
Thanks for the patch.  This seems like a reasonable addition, I'll merge later today.

PR #1668, opened 2020-01-03
Add: Heroku set up instructions for a rails app using sequel

BKSpurgeon:
### Why this PR?

* This PR documents instructions to help new users get started on sequel, rails and heroku. Hopefully it will make it easier for them.

I hope it helps.


jeremyevans:
After reviewing this, I think it contains too much information not specifically regarding Sequel (sequel-rails, fixture_dependencies, etc.).  As such, I don't think it is a good fit for inclusion in Sequel's documentation.  It would probably be better published as a blog post, or possibly added to sequel-rails documentation (if the sequel-rails maintainers want that).

PR #1672, opened 2020-01-15
Refactor when callbacks are fetched and calls inside remove_transactions

pabloh:
Refactor when callbacks are fetched and calls inside `remove_transactions`, in order to make implementing extensions easier.


jeremyevans:
This change looks fine to me, will merge shortly.

pabloh:
Awesome!

PR #1678, opened 2020-02-11
Call before_retry when retrying indefinitely

jcmfernandes:
The problem with conditionals and execution paths... not adding any spec to cover this feels wrong, but adding one tells more about the code than the intended behavior (feels like a regression test). I saw it failing by modifying https://github.com/uhoh-itsmaciek/sequel/blob/master/spec/core/database_spec.rb#L704-L706 adding the option `:num_retries => nil`. Feedback?

Taking the chance to thank you for this awesome piece of kit and all your other contributions to the ruby community 🙌 .

### DIFF
--- a/lib/sequel/database/transactions.rb
+++ b/lib/sequel/database/transactions.rb
@@ -185,12 +185,7 @@ module Sequel
           transaction(opts, &block)
         rescue *retry_on => e
           num_retries += 1
-          if tot_retries
-            if num_retries <= tot_retries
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch.  I agree that Sequel should be consistent here and call before_retry in both cases.

Please do add a spec for the behavior. The fact that it feels like a regression test is fine, as Sequel has lots of specs that are more like regression tests than actual specs.

jcmfernandes:
Added the spec.

PR #1683, opened 2020-03-30
Include source location of change block when raising error for irreversible migration

bethesque:
Thanks for all your work on this amazing gem Jeremy + other maintainers.

I put this little hack in to my own code, and thought it might be useful for someone else. I will not be offended if it is not thought to be worth merging!


jeremyevans:
Thanks for the patch.  This seems reasonable to me, I'll test and merge today and it should make the release on Wednesday.

PR #1684, opened 2020-04-01
Recognize timestamps with precision as timestamps

isc:
Since version 6, Rails generates timestamp columns with a specific precision (https://github.com/rails/rails/pull/34970).
At the moment Sequel doesn't recognise those columns as timestamp because of the added precision.

### DIFF
--- a/lib/sequel/database/query.rb
+++ b/lib/sequel/database/query.rb
@@ -325,7 +325,7 @@ module Sequel
         :integer
       when /\Adate\z/io
         :date
-      when /\A((small)?datetime|timestamp(\(\d\))?( with(out)? time zone)?)(\(\d+\))?\z/io
+      when /\A((small)?datetime|timestamp(\(\d\))?( with(out)? time zone)?)\z/io
         :datetime
       when /\Atime( with(out)? time zone)?\z/io
         :time
--- a/lib/sequel/database/query.rb
+++ b/lib/sequel/database/query.rb
@@ -325,7 +325,7 @@ module Sequel
         :integer
       when /\Adate\z/io
         :date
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch! I think it would be better to move the existing `\(\d+\))?` instead of adding a new one.  What do you think?

isc:
Maybe so, I don't know if it's the case that all database engines insert the precision part at this point in the type description. 
When you introduced the other `\(\d+\))?` 8 years ago here https://github.com/jeremyevans/sequel/commit/1f2cc837ff1e93b01e69c0e29fe4e46368628048, you mentioned Oracle and inserted it at the end. 
I don't have an Oracle instance at hand to verify but googling gives me this page https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch4datetime.htm#i1006050 which does contain `TIMESTAMP [(fractional_seconds_precision)] WITH TIME ZONE`, so maybe moving the existing `\(\d+\))?` is fine for Oracle.

jeremyevans:
Looking at 1f2cc837ff1e93b01e69c0e29fe4e46368628048, the commit message indicates support for `timestamp(N)`.  I think moving the `(\(\d+\))?` is probably safe.  If I'm wrong and there are databases that use `timestamp with time zone(N)`, we can add it back to the end.

jeremyevans:
Thanks for changing this.  I'll test and merge later today after the 5.31.0 release (this is too late to make the release as all prerelease testing has been completed).

jeremyevans:
Merged as 3c5036d1f9e3c425ed54b1690e26494bd5c99026

PR #1692, opened 2020-04-30
Add sequel-activerecord_connection and remove sequel-jsonapi_eager

janko:
This adds `sequel-activerecord_connection` gem to the list of external extensions.

It also removes `sequel-jsonapi_eager` gem because we've established that it's unsafe and that `tactical_eager_loading` plugin is better anyway (I've archived that repo already).


jeremyevans:
Thanks!

PR #1693, opened 2020-05-01
Mention subclasses plugin to freeze models & finalize their associations

qortex:
Following up on the google group discussion, simple doc PR to add a mention to the subclasses plugin where freezing is mentioned.

If you think it doesn't clutter this summary page.


jeremyevans:
Thanks for the patch!  This is a documentation only patch, so I should be able to get it in before the release today.

PR #1699, opened 2020-06-01
Replace abandoned gem sequel-savepoints-hooks

pabloh:
Replace `sequel-savepoints-hooks` gem with `sequel-force-hooks` since the former one has been abandoned


jeremyevans:
Is either gem needed at all now that `after_commit` and `after_rollback` support the `:savepoint` option?

pabloh:
Is not exactly the same use case, since with this force to run all the stored hooks force nested savepoints instead of declaring it inline like what you do with `after_commit: :savepoint`.

Is mostly useful when you have a transactional test suite and you, want to set a different place for the callbacks to run, without changing the code you want to test: https://github.com/pabloh/sequel-force-hooks#usage

jeremyevans:
I can't tell from the README if you can do more or less the same thing now with the run_transaction_hooks extension that ships with Sequel. Even if so, I don't have a problem updating the link, as the interface is different and some users may prefer it.

pabloh:
Ok

pabloh:
Thanks

jeremyevans:
Cherry-picked as ffa3c3dca4af5f50a0c121d8d95df810b374efdb

PR #1700, opened 2020-06-09
Fix link to bin/sequel guide

neko314:
Fix README, because `{bin/sequel guide}[rdoc-ref:doc/bin_sequel.rdoc] ` isn't converted to a link.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -73,7 +73,7 @@ Sequel includes an IRB console for quick access to databases (usually referred t
 
 You get an IRB session with the Sequel::Database object stored in DB.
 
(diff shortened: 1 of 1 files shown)


jeremyevans:
It is converted into a link, see http://sequel.jeremyevans.net/rdoc/files/README_rdoc.html#label-The+Sequel+Console

Now, GitHub doesn't correctly handle it, but that's a GitHub bug, not a bug in Sequel.

neko314:
I didn't understand rdoc syntax enough and I didn't know this is the matter of GitHub , but I got know now. Thank you for your review and reply!

PR #1703, opened 2020-07-14
Change code to code block in doc

arthurmmoreira:
Fix spaces to use code block instead of text on dataset filtering doc.


jeremyevans:
Thanks for the patch!

PR #1705, opened 2020-07-22
Add ORA-03135 as a connection error

sterlzbd:
Based on the causes given in the Oracle documentation I think `ORA-03135: connection lost contact` should be considered a connection error.

Causes from the oracle documentation (found [here](https://docs.oracle.com/cd/E11882_01/server.112/e17766/e2100.htm)):
1) Server unexpectedly terminated or was forced to terminate.
2) Server timed out the connection


jeremyevans:
Thanks for the patch! I agree with you, and will merge this shortly.

sterlzbd:
Great, thank you!

PR #1711, opened 2020-08-25
Don't eager load all result sets in ODBC adapter.

sparrovv:
At the moment ODBC connection will try to fetch all results and store
in memory before letting the caller access to any record.

This behaviour doesn't allow for streaming, makes some operation slower, and can lead to
OOM errors if someone tries to fetch a lot of records.


jeremyevans:
Thanks for the patch! This looks good, I'll try running the integration tests with it later today and if there are no issues, I will merge it.

jeremyevans:
No problems in the integration tests.  As I expected, this does break the case where you are trying to run queries inside a Dataset#each block, but that is considered undefined behavior and is an acceptable change.

sparrovv:
Thanks for a quick turnaround 🙇

PR #1713, opened 2020-08-28
Fix warning for last argument as keyword param

adam12:
Only appears when calling methods on the class and not the dataset.

```
lib/sequel/model/base.rb:755: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
spec/model/base_spec.rb:418: warning: The called method `with_foo' is defined here
```

### DIFF
--- a/lib/sequel/model/base.rb
+++ b/lib/sequel/model/base.rb
@@ -756,6 +756,7 @@ module Sequel
         else
           define_singleton_method(meth){|*args, &block| dataset.public_send(meth, *args, &block)}
         end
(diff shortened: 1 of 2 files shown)


jeremyevans:
This should use `ruby2_keywords(meth) if respond_to?(:ruby2_keywords, true)`.  And the spec needs to use eval so it works correctly (current code is a syntax error in Ruby <2.1).

jeremyevans:
Looks better.  I forgot to mention the spec should only be executed on Ruby 2.1+.  I would actually change to using a optional keyword argument with a default value that raises, then it will work on Ruby 2.0+.

jeremyevans:
Looks good.  I'll test and merge tomorrow.

PR #1717, opened 2020-09-16
Resolve warnings caused by undefined instance variables

bryanp:
I encountered a couple warnings originating in `sequel` when running a test suite with warnings enabled. These two commits resolve the warnings and best I can tell leave the underlying behavior unchanged.

### DIFF
--- a/lib/sequel/adapters/shared/postgres.rb
+++ b/lib/sequel/adapters/shared/postgres.rb
@@ -778,7 +778,7 @@ module Sequel
 
       # The version of the PostgreSQL server, used for determining capability.
       def server_version(server=nil)
-        return @server_version if @server_version
+        return @server_version if defined?(@server_version)
         ds = dataset
         ds = ds.server(server) if server
         @server_version ||= ds.with_sql("SELECT CAST(current_setting('server_version_num') AS integer) AS v").single_value rescue 0
--- a/lib/sequel/adapters/shared/sqlite.rb
+++ b/lib/sequel/adapters/shared/sqlite.rb
@@ -284,7 +284,7 @@ module Sequel
       end
 
       def begin_new_transaction(conn, opts)
-        mode = opts[:mode] || @transaction_mode
+        mode = opts[:mode] || transaction_mode
         sql = TRANSACTION_MODE[mode] or raise Error, "transaction :mode must be one of: :deferred, :immediate, :exclusive, nil"
         log_connection_execute(conn, sql)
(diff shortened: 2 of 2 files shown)


jeremyevans:
Sequel by design does not initialize instance variables to `nil` or use `defined?` checks everywhere instance variables are accessed.  There are far more places in Sequel where you can run into uninitialized instance variable warnings than these two cases.  Starting in Ruby 2.4, you can override `Warning.warn` (e.g. via the warning gem) to filter these warnings.

I've been trying to get ruby-core to accept per-object handling of uninitialized instance variable warnings, so that I could turn off these pointless verbose-mode warnings in Sequel, but so far, no luck.

So far, I've been unwilling to ship an extension with Sequel that you could use that would automatically filter the warnings, but I will leave this open while I give that another thought.

bryanp:
I use the `warning` gem to filter out warnings on a per-gem basis, but wanted to contribute these fixes before ignoring ones that originate in `sequel`. If your policy is to leave the uninitialized instance variable warnings unhandled that's fine.

There are several places that use `defined?` in memoized methods (e.g. [adapters/shared/postgres.rb:819-822](https://github.com/jeremyevans/sequel/blob/ff5d77cb60a61b41d3eb500344f287f0b9fbdb97/lib/sequel/adapters/shared/postgres.rb#L819-L822)) and/or call the reader instead of accessing the instance variable directly. From that perspective, I think these changes could be seen as bringing consistency to the codebase while also resolving a couple of warnings.

jeremyevans:
It is Sequel's policy to not initialize instance variables to `nil` for the sole sake of avoiding uninitialized instance variable warnings.

`defined?` is used in places like your example where we need to differentiate between non-initialized and explicitly set to `nil`/`false`.  We don't want to be issuing a query every time `supports_prepared_transactions?` is called if they aren't supported. In places where we don't need to differentiate, accessing the instance variable directly is generally done as it is faster.

bryanp:
Ah, I see. Thank you for the clear explanation.

jeremyevans:
I've given this some more thought and still don't think it's worth shipping an extension for this, when you can handle it via:

```ruby
require 'warning'
Warning.ignore(:missing_ivar,  Gem.loaded_specs['sequel'].full_gem_path)
```

However, if there is sufficient demand for it, I will reconsider.  Please open a thread on the sequel-talk Google Group to gauge community desire for such an extension if you would want it added.

PR #1719, opened 2020-09-27
Some doc fixes

janko:
I believe one of the savepoint paragraphs was accidentally duplicated, as the same scenario is described and shown right above it.

I also addressed two typos and an early newline in comments.


jeremyevans:
Looks good, thanks!

jeremyevans:
Merged at 92d2b0b866cdab03aa7d365d3573f46198545b99

PR #1720, opened 2020-09-28
Support unsigned decimals

akimd:
Hi,

This PR addresses a limitation I hit: one of my columns' type was "decimal(7,2) unsigned", which was not recognized and therefore treated as a string.

Cheers!

### DIFF
--- a/lib/sequel/extensions/schema_dumper.rb
+++ b/lib/sequel/extensions/schema_dumper.rb
@@ -37,7 +37,7 @@ module Sequel
         {:type =>schema[:type] == :boolean ? TrueClass : Integer}
       when /\Abigint(?:\((?:\d+)\))?(?: unsigned)?\z/
         {:type=>:Bignum}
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the report and patch.  Can you refactor your patch to not use the Regexp `x` modifier?

akimd:
Sure.  So you don't think it makes things clearer?

jeremyevans:
I would prefer that we not add it, for consistency with the other regexps.  Also, I don't think it makes things clearer.

akimd:
Jeremy, do you have a release scheduled?  I would need to update that gem (and tapsoob, which depends upon it) in my production deployments.  TIA!

flash-gordon:
@akimd see https://github.com/jeremyevans/sequel#sequel-release-policy- and https://github.com/jeremyevans/sequel/releases I would expect the next release on Oct 1 or so :)

akimd:
> https://github.com/jeremyevans/sequel/releases

Oh yes, I sense there's a pattern in there :-)

PR #1723, opened 2020-10-05
Remove duplicated comment line

dentarg:
I guess it was accidentally added in d592153


jeremyevans:
Thanks for the patch!

PR #1725, opened 2020-10-13
Remove redundant #freeze calls

the-spectator:
## Summary
This PR removes redundant `#freeze` calls on string literals. The calls are redundant as we are using `# frozen_string_literal: true`.

NOTE: If this PR feels more of a cosmetic change, please do close it. Thank you!


jeremyevans:
Sequel currently supports Ruby 1.9+, and the frozen_string_literal magic comment is only respected in Ruby 2.3+.  If we drop support for Ruby versions before 2.3, then we could apply this.  However, as there are no current plans to do that, I'm going to close this.

PR #1735, opened 2020-11-11
Fix CsvSerializer-association_join compatibility

rlue:
Prior to this commit, the CsvSerializer plugin
converted model instances (records) into CSV rows
using the getter methods defined on the model class
corresponding to its DB table's attributes / column names.

This approach fails when combining #to_csv with #association_join,
because the latter can introduce unexpected attributes
for which no getter methods exist. For instance,

    Album
      .select(Sequel[:artist][:name].as(:artist_name), :name)
      .association_join(:artist)
      .where(name: /california/i)
      .to_csv

raises

    NoMethodError: undefined method `artist_name' for #<Album:0x0000563a50e013f8>
    Did you mean? artist
    from /path/to/sequel-5.38.0/lib/sequel/plugins/csv_serializer.rb:189:in `public_send'

This commit adopts `Model#[]` over `#public_send` to avoid this problem.

### DIFF
--- a/lib/sequel/plugins/csv_serializer.rb
+++ b/lib/sequel/plugins/csv_serializer.rb
@@ -186,7 +186,7 @@ module Sequel
 
           CsvSerializer.csv_call(:generate, opts) do |csv|
             items.each do |object|
-              csv << headers.map{|header| object.public_send(header)}
+              csv << headers.map{|header| object[header]}
             end
           end
         end


rlue:
This PR is a follow-up to a problem I encountered related to [this discussion](https://groups.google.com/g/sequel-talk/c/vvVenEK1aOk) on the sequel-talk Google Group.

Not sure if there are any gotchas or hidden dangers to this fix, but **caveat—it introduces an inconsistency between [L169 and L189](https://github.com/jeremyevans/sequel/blob/master/lib/sequel/plugins/csv_serializer.rb#L169-L189):**

```ruby
169 csv << headers.map{|k| public_send(k)}
189 csv << headers.map{|header| object.public_send(header)}  # becomes object[header]
```

Changing L189 didn't break any tests, but changing L169 broke three, so this commit leaves L169 alone. Tried to write a unit test for it, but I currently don't understand the CsvSerializer well enough to do so. :\

Happy to take a different approach if you think this one isn't right.

jeremyevans:
The problem with this approach is that it will break code for people who override model instance methods.  I believe the other serializers also call methods instead of using `#[]`, so this would make the CSV serializer inconsistent.

I think the correct solution in this case is to do to one of the following:

1. Create the CSV manually instead of using Model.to_csv

2. Use def_column_accessor to create accessor methods for the joined columns

3. Add an option for using `#[]` instead of methods, so that the behavior is backwards compatible by default.

rlue:
> The problem with this approach is that it will break code for people who override model instance methods.

That makes sense. 

> I believe the other serializers also call methods instead of using #[], so this would make the CSV serializer inconsistent.

Hm, that means the other serializers also suffer from the same problem then. What would you think of the following on `Sequel::Model` instances?

```ruby
def method_missing(m, *args, &block)
  super unless values.key?(m)
  raise ArgumentError, <<~MSG.chomp if args.any?
    wrong number of arguments (given #{args.length}, expected 0)
  MSG

  values[m]
end

def respond_to_missing?(m, include_private)
  values.key?(m) || super
end
```

I would have modified this PR, but I'm honestly not sure where to put it. Could it go in `Sequel::Model::InstanceMethods`?

jeremyevans:
No, I'm definitely against a method_missing solution.  However, you are definitely free to use that solution in your application, it sounds like it would meet your needs.

rlue:
Okay, cool. (Mind if I ask why? Just trying to learn!)

> you are definitely free to use that solution in your application

Will do. In the meantime, I'd still be interested in solving this problem generally for sequel, provided you are open to it. Note that:

1. It's not just joined columns; it's also aliased columns, too:

   ```ruby
   >> Album.select(Sequel.as(:name, :title)).to_csv
   NoMethodError: undefined method `title' for #<Album @values={:title=>"Now That's What I Call Music! 17"}>
   ```

2. Upon closer inspection, it appears that JsonSerializer also suffers from this issue, but XmlSerializer does not. That is, XmlSerializer uses `#[]` and not `#public_send(column_name)`; _i.e.,_ it does not honor user-defined / overridden getter methods.

This means that the behavior of existing serializers is not consistent, and making them so out of the box would be a breaking change (and thus, presumably, would have to wait until the next major version bump).

What do you think?

jeremyevans:
As a general rule, I avoid `method_missing` unless it is absolutely required for the use case (e.g. Sequel virtual rows).  In my opinion, `method_missing` leads to code that is more difficult to understand, and using it is always a tradeoff.

In terms of changes to Sequel, as I said earlier I'm open to an option to `to_csv` to use `#[]` instead of `public_send`.  That is backwards compatible, works as expected when instance methods are overridden, and allows handling of this use case.

For json_serializer, I'm not sure this is needed.  With json serializer, you don't generally join, you generally use associations and include nested associations.  CSV can't handle nesting, so that is a different case.  I'm not necessary opposed to a similar solution for json_serializer, but in general we should wait to see if it is actually needed.

rlue:
> in general we should wait to see if it is actually needed.

I presume the same logic applies for changing XmlSerializer to behave consistently with its CSV & JSON counterparts?

Based on your response, I'm perfectly happy to close this PR. I thought it might be nice to make the different serializers work consistently and for certain edge cases not to blow up out of the box, but I see how that conflicts with other priorities for the project, like code legibility and backwards compatibility.

Thanks again for hashing it out with me!

PR #1741, opened 2020-11-30
Remove ".sql" call in cheat sheet examples

jlmuir:
Since the cheat sheet uses the convention that the resulting SQL is
given in an adjacent comment, including the `.sql` method call in the
examples is unnecessary.  Also, sometimes the `.sql` method call was
included, and sometimes not.  Remove all occurrences of it except, of
course, in the Retrieving SQL examples.

### DIFF
--- a/doc/cheat_sheet.rdoc
+++ b/doc/cheat_sheet.rdoc
@@ -95,18 +95,18 @@ Without a filename argument, the sqlite adapter will setup a new sqlite database
 
 === AND/OR/NOT
 
-  DB[:items].where{(x > 5) & (y > 10)}.sql 
+  DB[:items].where{(x > 5) & (y > 10)} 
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch!   For future documentation fixes, can you please submit them in a single pull request after all fixes have been made, instead of separate pull requests for each file?

jeremyevans:
Cherry-picked at de98fc703cdc53ec92dd33ca611873d2d010dccf

jlmuir:
> Thanks for the patch! For future documentation fixes, can you please submit them in a single pull request after all fixes have been made, instead of separate pull requests for each file?

Sure.  I'm mostly just finding these as I'm reading, but I can buffer them and only submit when I'm unlikely to be reading more.

PR #1757, opened 2021-04-09
Fix typo in pg_enum documentation

celsworth:
Trivial documentation fix I spotted today.


jeremyevans:
Thanks for the patch!

PR #1760, opened 2021-04-22
Mention :include in add_index docs (PostgreSQL)

smaximov:
Hello! I wanted to create a covering index for an existing PostgreSQL table, but the docs don't mention this option in `AlterTableGenerator#add_index` (`CreateTableGenerator#index` docs have it, though, so I copy-pasted it).


jeremyevans:
Thanks for the patch! Looking at the existing documentation for the index methods, it is fairly inconsistent.  I'd like to fix this a different way, by having `AlterTableGenerator#add_index` reference `CreateTableGenerator#index` so we don't need to duplicate all of the documenation.  I'll make that change shortly.

PR #1769, opened 2021-06-10
Add ability to dump UUID columns

michalg-:
Add ability to dump UUID columns using SchemaDumper
migration file:
```ruby
# frozen_string_literal: true
Sequel.migration do
  up do
    run 'CREATE EXTENSION "uuid-ossp"'
    create_table(:users) do
      column :id, :uuid, default: Sequel.function(:uuid_generate_v4), primary_key: true
    end
  end

  down do
    drop_table(:users)
    run 'DROP EXTENSION IF EXISTS "uuid-ossp"'
  end
end

```

before:
```ruby
Sequel.migration do
  change do
    create_table(:schema_migrations) do
      String :filename, :text=>true, :null=>false
      
      primary_key [:filename]
    end
    
    create_table(:users) do
      String :id, :null=>false
      
      primary_key [:id]
    end
  end
end

```

after:
```ruby
Sequel.migration do
  change do
    create_table(:schema_migrations) do
      String :filename, :text=>true, :null=>false
      
      primary_key [:filename]
    end
    
    create_table(:users) do
      column :id, :uuid, :default=>Sequel::SQL::Function.new!(:uuid_generate_v4, [], {}), :null=>false
      
      primary_key [:id]
    end
  end
end

```

### DIFF
--- a/lib/sequel/database/query.rb
+++ b/lib/sequel/database/query.rb
@@ -12,12 +12,12 @@ module Sequel
 
     # The prepared statement object hash for this database, keyed by name symbol
     attr_reader :prepared_statements
-    
+
     # Whether the schema should be cached for this database.  True by default
     # for performance, can be set to false to always issue a database query to
     # get the schema.
     attr_accessor :cache_schema
-    
+
     # Runs the supplied SQL statement string on the database server.
     # Returns self so it can be safely chained:
     #
@@ -26,7 +26,7 @@ module Sequel
       run(sql)
       self
     end
-    
+
     # Call the prepared statement with the given name with the given hash
     # of arguments.
     #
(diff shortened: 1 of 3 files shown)


jeremyevans:
I appreciate that you took the time to send in a patch. Unfortunately, I cannot accept this, for the following reasons:

* The change in `column_schema_default_to_ruby_value` returns an SQL expression and not a value, and it must return a value.  Additionally, it makes the assumption that any function that contains `uuid_` will have no arguments, when there is no guarantee that is the case.

* The change to `column_schema_to_ruby_type` uses a specific type of `:uuid`.  This method is used for the generic schema dumper, and should only use generic types that Sequel can translate into appropriate database-specific types.

* Your code mixes whitespace changes with code changes, which is explicitly warned against in the contribution guidelines.

I think the main problem is that you are trying to support uuid as a generic type, when the type is PostgreSQL-specific.  Since you are using a PostgreSQL-specific type, you should use the database-specific schema dumper instead of the generic schema dumper. With bin/sequel, you should use `-D` instead of `-d`.  If you are using the schema dumper directly, you should use the `:same_db` option when dumping the schema.

michalg-:
Thanks for the answer, I see your point of view. Will check those options you suggest. Have a nice day!

PR #1771, opened 2021-06-29
Check if server_version is defined

graywolf-at-work:
If running ruby with warnings enabled, the postgresql adapter complains that:

/home/showmax/bundle/ruby/2.7.0/gems/sequel-5.45.0/lib/sequel/adapters/shared/postgres.rb:781: warning: instance variable @server_version not initialized

Adding defined? into the condition resolved the warning.

### DIFF
--- a/lib/sequel/adapters/shared/postgres.rb
+++ b/lib/sequel/adapters/shared/postgres.rb
@@ -778,7 +778,7 @@ module Sequel
 
       # The version of the PostgreSQL server, used for determining capability.
       def server_version(server=nil)
-        return @server_version if @server_version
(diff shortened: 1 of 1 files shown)


jeremyevans:
This is expected behavior and not a bug. Sequel uses uninitialized instance variables in many places, not just this one.  Thankfully, ruby-core recognized they aren't actually a problem and eliminated the verbose warning in 3.0.  In 2.7 you can use the warning gem to easily filter out these pointless warnings.

graywolf-at-work:
> eliminated the verbose warning in 3.0

I did not know this, thanks :)

PR #1773, opened 2021-07-08
Fix grammar in association_basics.

plujon:


jeremyevans:
Thanks for the patch!  All of these changes look good.

PR #1774, opened 2021-07-12
Support renaming multiple tables at once in MySQL

nick96:
MySQL has the `RENAME TABLE` statement that allows us to atomically
rename multiple tables. This doesn't appear to be supported in other
RDBMS's so I've only added it to the database methods for the MySQL
adapters (as recommended when I posted in the Google Group).


jeremyevans:
Looks good.  I'll test and merge tomorrow.

jeremyevans:
Merged at 91a7099732c05cb3fc3f4aa18dec18caa3bdcd53

PR #1775, opened 2021-07-18
Fixing timestamp code example

RafaelFerraro:


lib/sequel/plugins/timestamps.rb
```diff
@@ -19,7 +19,7 @@ module Plugins
     #
     #   # Timestamp Artist instances, forcing an overwrite of the create
```
RafaelFerraro, line 20: The description is referring to an _Artist_ model, but on the code example it's using an _Album_ model. Just a small fix.

jeremyevans:

PR #1777, opened 2021-07-28
Fix example for `:dataset` in associations doc

v-kolesnikov:
I'm not sure 100%, but looks like there is a typo in example for `:dataset` option.


jeremyevans:
Yep, that's definitely a typo.  Thanks for the patch!

PR #1780, opened 2021-08-16
docs | fix typo OR in place of AND

Vulwsztyn:


jeremyevans:
Thanks for the patch!

PR #1791, opened 2021-11-14
Add an :extra key for rows generated from MySQL's schema

bschmeck:
The value of this field is the unparsed output of whatever was returned by MySQL's DESCRIBE statement in the Extra column.


jeremyevans:
This looks good.  I'll test and merge on Monday.

bschmeck:
Thanks!

PR #1812, opened 2021-12-24
Use ‘{}’ instead of ARRAY[] when array is empty in postgres

hieuk09:
I have a few columns with array type in my postgres DB. However, if I insert to those columns using empty array, the result SQL is invalid:

```ruby
DB.create_table :items do
  column :id, Integer, primary: true
  column :tags, 'text[]', null: false
end

DB[:items].insert(id: 1, tags: Sequel.pg_array([]))
# raise Sequel::DatabaseError: PG::IndeterminateDatatype: ERROR:  cannot determine type of empty array
# LINE 1: INSERT INTO "items" ("id", "tags") VALUES (1, ARRAY[]) RETUR...
#                                                       ^
# HINT:  Explicitly cast to the desired type, for example ARRAY[]::integer[].
```

This can be worked around by adding type to `pg_array` like this:

```ruby
DB[:items].insert(id: 1, tags: Sequel.pg_array([], 'text'))
```

However, that's quite tedious for big application to create mapping for all array columns. Since Postgres supports empty array literal `'{}'`, if Sequel could convert to that when no type is passed in, that would be great.

## Testing

After applying this fix, the original code works

```ruby
DB.create_table :items do
  column :id, Integer, primary: true
  column :tags, 'text[]', null: false
end

DB[:items].insert(id: 1, tags: Sequel.pg_array([]))
=> nil
```

I also run my applications with this patch and everything seems ok (the test suite passed, most functionality works). However, I've only tested with posgres 13 & 14.

Please let me know if there are problems with this PR.

### DIFF
--- a/lib/sequel/extensions/pg_array.rb
+++ b/lib/sequel/extensions/pg_array.rb
@@ -446,7 +446,7 @@ module Sequel
       # database array type.
       def sql_literal_append(ds, sql)
         at = array_type
-        if empty? && at
+        if empty?
           sql << "'{}'"
         else
           sql << "ARRAY"
--- a/spec/extensions/pg_array_associations_spec.rb
+++ b/spec/extensions/pg_array_associations_spec.rb
@@ -648,7 +648,7 @@ describe Sequel::Model, "pg_array_associations" do
     @o1.tag_ids.must_equal []
     @db.sqls.must_equal []
     @o1.save_changes
-    @db.sqls.must_equal ["UPDATE artists SET tag_ids = ARRAY[] WHERE (id = 1)"]
+    @db.sqls.must_equal ["UPDATE artists SET tag_ids = '{}' WHERE (id = 1)"]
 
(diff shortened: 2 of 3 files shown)


jeremyevans:
The problem with this approach is the expression will be treated as an unknown type, which will get autocasted to anything.  Constructions that should raise an error stop raising an error.  For example, `DB.get(Sequel[""] + Sequel.pg_array([]))` should raise an exception, because you cannot add an array to a string, but with this change, that will stop raising an exception. In other words, this change reduces misuse resistance.

I think the better approach here is to explicitly use the correct array type, even if it is considered tedious.  Are there cases were you cannot determine the correct array type to use?

If you work with models, this is less of an issue, because the setter methods in the model will automatically cast arrays to the expected type.

hieuk09:
Thank you for the detail explanation, that's very insightful and I totally agree with your decision here.

PR #1830, opened 2022-01-25
Fix doc for pg array any

artofhuman:


jeremyevans:
Thanks for the patch!

PR #1870, opened 2022-05-13
Make create_function of the PostgreSQL-adapter accept the PARALLEL pa…

bananarne:
…rameter

https://www.postgresql.org/docs/14/sql-createfunction.html states a parallel parameter/configuration option which comes available through this.


jeremyevans:
Thanks for the patch!  I'll cherry-pick this shortly.

PR #1872, opened 2022-05-17
Use https in site search form

krelly:


jeremyevans:
Thanks for the patch!

krelly:
Wow that was fast 😀

PR #1882, opened 2022-06-07
Fix typo in `pg_hstore` docs

dentarg:
Typo introduced in https://github.com/jeremyevans/sequel/commit/60795802a37c680c76d3aff33aa118bd3994aa0e but the other case was already fixed by https://github.com/jeremyevans/sequel/commit/697f2ed77015518b75ee4c36f686e0bc155f745a


jeremyevans:
Thanks for the patch!

PR #1899, opened 2022-07-29
Update plugins.html.erb

kwong-yw:
Adds a reference to the `sequel-snowflake` adapter, to be used with Snowflake, the cloud-based database service.


jeremyevans:
Thanks for the patch!

PR #1910, opened 2022-08-21
Implements Sequel::Dataset#not_exists

kenaniah:
Adds a method for generating `NOT EXISTS` clauses within a dataset.

### DIFF
--- a/lib/sequel/dataset/sql.rb
+++ b/lib/sequel/dataset/sql.rb
@@ -14,6 +14,14 @@ module Sequel
     def exists
       SQL::PlaceholderLiteralString.new(EXISTS, [self], true)
     end
+
+    # Returns a NOT EXISTS clause for the dataset as an SQL::PlaceholderLiteralString.
+    #
+    #   DB.select(1).where(DB[:items].not_exists)
+    #   # SELECT 1 WHERE (NOT EXISTS (SELECT * FROM items))
+    def not_exists
+      SQL::PlaceholderLiteralString.new(NOT_EXISTS, [self], true)
+    end
     
     # Returns an INSERT SQL query string.  See +insert+.
     #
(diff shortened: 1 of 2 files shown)


jeremyevans:
You can already use `DB[:b].exclude(DB[:a].exists)` or `DB[:b].where(~DB[:a].exists)` to get the same behavior.  I don't think it is worth adding a separate `not_exists` method.  However, I'll leave this open for feedback for a few days.

AlexWayfer:
It's not actually the same, but as I know we don't have something like `Sequel::NOT_NULL`, we use `exclude(a: nil)` or `~`, so I can agree with @jeremyevans. Commonly I'm not against sugar and variety, but this one seems not enough profitable. Maybe only if you have specific use-cases?

kenaniah:
I believe that one of the major benefits of having an explicit `#not_exists` defined is that it improves both readability and searchability within the end user's codebase. Consider a global search across a large codebase -- if all you have is `#exists`, it would be challenging to immediately identify the intent of a given statement, as the inversion may not be local (either `~` used on a different line, `#exclude` vs `#where`, etc.). 

Furthermore, I do believe that `#not_exists` would improve the usability of Sequel as a library by lowering the cognitive load of determining how to implement a `NOT EXISTS` filter, especially given that the overloaded binary ones complement operator (`~`) is probably not the first thing a user is going to reach for.

AlexWayfer:
> Consider a global search across a large codebase -- if all you have is `#exists`, it would be challenging to immediately identify the intent of a given statement, as the inversion may not be local (either `~` used on a different line, `#exclude` vs `#where`, etc.).

Honestly — when you have around 100 matches and 15 on the screen, it doesn't matter, look for `not_` or `~` before `exists`. But maybe you have more specific use-case.

jeremyevans:
I think the vast majority of usage would be `where(dataset.not_exists)`, which I think is better served by `exclude(dataset.exists)`.  For the more exotic usage, `~dataset.exists` seems better than `dataset.not_exists`, because it shows how `~` works as a generic inversion operator in Sequel.  In general, I think it is better to use existing generic approaches, as opposed to adding new specific approaches.

Weighing the pros and cons, I don't think it's worth adding this to Sequel.  This could easily be shipped as an external extension for the users who want it.

PR #1925, opened 2022-09-17
Allow up/down in migrations to take SQL directly

will:
Rather than needing to do

```ruby
up do
  execute <<~SQL
    create table ...
  SQL
end
```

with this patch you can pass in a sql string directly

```ruby
up <<~SQL
  create table ...
SQL
```

### DIFF
--- a/lib/sequel/extensions/migration.rb
+++ b/lib/sequel/extensions/migration.rb
@@ -129,7 +129,11 @@ module Sequel
     end
 
     # Defines the migration's down action.
-    def down(&block)
+    def down(raw_sql = nil, &block)
+      if raw_sql
+        Kernel.raise(ArgumentError, "down cannot be given both raw sql and a block") if block
+        block = Proc.new { execute raw_sql }
+      end
       migration.down = block
     end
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the pull request.  This is an interesting idea, but I'm leaning against it. If multiple SQL statements in a single string was supported by Sequel (it isn't, though it does work on some adapters), I could see supporting this.  However, since Sequel only officially supports a single SQL statement per string, this limits you to a single SQL statement per up/down migration.  I think that's too limiting, so to me the costs seem to outweigh the benefits at this point.  However, I'm open to feedback regarding this.

If we do accept this, this should use `run` and not `execute`, as `execute` is an adapter-specific API, and limited to `SELECT` queries in some adapters.

jeremyevans:
After giving this some thought, I don't think the benefits outweigh the costs.  It would make some things easier, but considering it's only a few more characters to get the same behavior currently, it doesn't seem worth it.  If this is for something that is used all the time, the character savings would mean more, but migration code should be a fairly small percentage of all code in any reasonable application, and this only saves about 10 characters per migration.

PR #1937, opened 2022-10-06
fix minor typo (?): "emualting" vs "emulating"

benkoshy:
IDK: looks like it should be "emulating"?


jeremyevans:
Thanks for the patch!

PR #1941, opened 2022-10-12
add: docs - virtual row example for reversing

benkoshy:
### Why?

* Present a different option to do the same thing.
* Implicitly teaches the DSL with simple queries + shows a model example: `Artist`

### Testing

`bin/sequel -c  "puts DB[:artists].order{[name, id.desc]}.sql"`

`SELECT * FROM artists ORDER BY name, id DESC`

Pls LMK if this hurts the docs, rather than improves it.

### DIFF
--- a/doc/virtual_rows.rdoc
+++ b/doc/virtual_rows.rdoc
@@ -166,7 +166,7 @@ distinct method on the returned Function:
   ds.select{|o| o.count(o.col1, o.col2).distinct}
   ds.select{count(col1, col2).distinct}
   # SELECT count(DISTINCT col1, col2)
-  
+
 == SQL::Functions with windows - SQL window function calls
 
 To create a window function call, just call the over method on the Function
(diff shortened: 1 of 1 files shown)


jeremyevans:
The querying guide does mention virtual row blocks earlier.  I don't think it makes sense to discuss virtual row blocks here.  You could probably have something like this in the virtual row section, and mention that you can use virtual row blocks in many dataset methods, not just filters.

benkoshy:
Pls revert if inadequate in anyway. I hope it is helpful.

jeremyevans:
At the top of the virtual row documentation it says:

```
Dataset methods where, order, and select all take blocks that are referred to as
virtual row blocks.  Many other dataset methods pass the blocks
they are given into one of those three methods, so there are actually
many Sequel::Dataset methods that take virtual row blocks.
```

I don't think this example adds value, so I'm going to close this.

PR #1944, opened 2022-10-14
Add: Heroku set up instructions for a rails app using sequel

christianfrstorm:
Great


jeremyevans:
Sorry, but I don't think this is a good fit for Sequel's documentation, especially considering Heroku is dropping their free plans next month.  I encourage you to host this information elsewhere.

PR #1946, opened 2022-10-19
Update virtual_rows.rdoc - add emphasis

benkoshy:
Add: emphasis. 

### Why?
I had to do a double-take.

Pls LMK if like  proposals are not be well received. I make them in the hope they will be useful. but they do incur a small maintenance overhead.

### DIFF
--- a/doc/virtual_rows.rdoc
+++ b/doc/virtual_rows.rdoc
@@ -23,7 +23,7 @@ With virtual rows, you can use the less verbose:
 Virtual row blocks behave differently depending on whether the block accepts
 an argument.  If the block accepts an argument, it is called with an instance
 of Sequel::SQL::VirtualRow.  If it does not accept an argument, it is
-evaluated in the context of an instance of Sequel::SQL::VirtualRow.
+evaluated in the <em> context of an instance </em> of Sequel::SQL::VirtualRow.
 
   ds = DB[:items]
   # Regular block


jeremyevans:
Thanks for submitting a pull request.  I'm not sure why the extra emphasis is beneficial in the is case, can you explain what value it adds?

benkoshy:
>  I'm not sure why the extra emphasis is beneficial in the is case, can you explain what value it adds?

This was how _**I read**_ the docs:

> If the block accepts an argument, it is called with an instance of [Sequel::SQL::VirtualRow](https://sequel.jeremyevans.net/rdoc/classes/Sequel/SQL/VirtualRow.html). 
> If it does not accept an argument, it is called with an instance of [Sequel::SQL::VirtualRow](https://sequel.jeremyevans.net/rdoc/classes/Sequel/SQL/VirtualRow.html).

......wait, what? They are the same whether they are with an argument or without? That can't be right. Let's do a double-take and read it again, a little closer:

If the block accepts an argument, it is called with an instance of [Sequel::SQL::VirtualRow](https://sequel.jeremyevans.net/rdoc/classes/Sequel/SQL/VirtualRow.html). If it does not accept an argument, it is **EVALUATED IN THE CONTEXT** of an instance of [Sequel::SQL::VirtualRow](https://sequel.jeremyevans.net/rdoc/classes/Sequel/SQL/VirtualRow.html).

There **_is_** a difference! But it ought to be highlighted - otherwise the nuance is easy to miss - to me at least.

my two cents.

jeremyevans:
Thanks for the explanation.  I'll merge this shortly.

jeremyevans:
Cherry-picked as 1c46114eb2d571522765f843f974dc4187d1e71b

PR #1948, opened 2022-10-21
Use Ruby 1.9+ Hash syntax in all of the docs

janko:
Most of the Sequel docs already use the modern Hash syntax, so this PR tries to convert all the remaining places.


jeremyevans:
Looks good, thanks for working on this!

PR #1950, opened 2022-10-24
Improve error message when migrator is not current

janko:
Given that `Sequel::Migrator::NotCurrentError` exception class name already contains the information "migrator is not current", we could use the exception message to provide more details, so that it's clearer to the developer what the error means (if they don't know about checking of current migration).


jeremyevans:
Thanks for the patch!

jeremyevans:
Cherry-picked at 851046a38b832ebf9cf435a2290debade184dd90

PR #1951, opened 2022-10-24
Add advice for making migrations work with code reloading

janko:
When I was setting up a Roda & Sequel application with Zeitwerk, I was pondering on how to solve syncing models after running migrations in development. One idea was to disable schema caching globally altogether in development, at the cost of more verbose SQL logs and performance. Another idea was to somehow clear the schema cache when model classes are unloaded, but I couldn't find any public methods that do that.

It's only when I looked at the roda-sequel-stack project that I discovered the `cache_associations` config, which the docs say is recommended to disable when using code reloading. It's perfect, because it doesn't require hooking into code reloading. And reading the Rails source code gave me the idea to trigger the reload when the schema file changes.

Since it took me a while to arrive to these strategies, I thought it would be useful to include them in the migration guide. For those curious, this is how I implemented it in my project (the Listen gem doesn't support listening on individual files, so it listens on the whole `db/` directory):

```rb
# config/loader.rb
loader = Zeitwerk::Loader.new
loader.enable_reloading if ENV["RACK_ENV"] == "development"
# ...

if ENV["RACK_ENV"] == "development"
  listener = Listen.to("app", "db") { loader.reload }
  listener.start
end
```
```rb
# config/initializers/sequel.rb
# ...
Sequel::Model.cache_associations = false if ENV["RACK_ENV"] == "development"
```

### DIFF
--- a/doc/migration.rdoc
+++ b/doc/migration.rdoc
@@ -643,3 +643,11 @@ If you need to use database extensions in migrations (e.g. +:pg_enum+), you shou
       # migration here
     end
   end
+
+== Code Reloading
+
+When using code reloading in development, you need to make sure that Sequel models are reloaded when new migrations are applied, so that they pick up schema changes such as newly added or removed columns. One way you could do that is ensure schema is dumped when running the migrations, and trigger a reload when the schema file has changed.
+
+You'll also want to configure Sequel models to reload the schema cache when they're re-defined, to ensure current schema is picked up:
+
+  Sequel::Model.cache_associations = false if ENV["RACK_ENV"] == "development"


AlexWayfer:
> I'm not sure I understand this. Changes in migration files will not do anything by default, migrations will run only when the developer runs them. Ideally you'd listen just on `db/schema.{rb,sql}` changes, but Listen gem can only listen on directories, it seems.

OK, sorry, probably I don't understand the problem.

You're listening to the migrations directory. You create a new migration here (or change an existing one). And… what next? What do you expect? Like, you have running server, you run migrations in a separate terminal and still running server uses a new schema? Or what exactly? Because I see no problem to stop the server, run migrations (or rollback and re-run) and start in the same term.

janko:
> You're listening to the migrations directory. You create a new migration here (or change an existing one). And… what next? What do you expect? Like, you have running server, you run migrations in a separate terminal and still running server uses a new schema? Or what exactly? Because I see no problem to stop the server, run migrations (or rollback and re-run) and start in the same term.

Yes, when I run the migrations, that changes the schema, so the schema file changes, which fires of the listener callback that triggers a code reload, and when models get reloaded so does the schema cache (provided that `cache_associations` is set to `false`). So, you're able to run migrations up or down without restarting the server.

AlexWayfer:
@janko then I'd change the text of proposing changes to "hot code reloading", because just "code reloading" can be done via server restart, there is difference, example of discussion: https://stackoverflow.com/a/41429055

And still it's very specific case for me, and I'm not sure it worth to be mentioned in official docs, especially… you suggest to disable associations cache, what about new models? What about old models (drop table)? What if it's working now correctly, but sometimes will be changed (cache method of models, or their columns/methods, especially with plugins like enum, not only associations).

jeremyevans:
The language and placement in the documentation implies you need `cache_associations` set to false after running a new migration, but the reality is you should have it set to false any time you are using code reloading, even if not running migrations.  Otherwise associations will cache classes on first access.

I don't think the difference matters when running Zeitwerk, since Zeitwerk will drop all cached constants when reloading.  However, it matters when running rack-unreloader, which only reloads constants for files that change (a much faster approach, at least for large applications).

Can you move this to the Association Basics guide, and just have it state that you should have `Sequel::Model.cache_associations = false` when doing any code reloading that doesn't involve restarting the related process?

janko:
OK, since this would be a different change, one focused on making association reflections work when partially reloading code, I will open a separate pull request.

The primary goal behind this pull request was to set up automatic schema reloading when models are reloaded, so that they pick up new schema changes. I used disabling `Sequel::Model.cache_associations` for its secondary effect that it changes `Sequel::Model.reload_db_schema?` to return `true`, which was the only way I found how to do that. As you said, when using Zeitwerk, caching association reflections isn't problematic, as all model classes are always reloaded.

> I'd change the text of proposing changes to "hot code reloading", because just "code reloading" can be done via server restart, there is difference, example of discussion: https://stackoverflow.com/a/41429055

@AlexWayfer I couldn't find any code reloader listed in [roda docs](https://roda.jeremyevans.net/rdoc/files/README_rdoc.html#label-Code+Reloading) mentioning they're doing "hot code reloading", I only heard that term being used in JavaScript and I believe Elixir. I think Rubyists just use "restart" for what that StackOverflow answer calls "live reload", and I would prefer to keep using the existing terminology.

PR #1952, opened 2022-10-24
Add sequel_tools gem to list of external extensions

janko:
I found it really convenient for quickly setting up Rake tasks for managing the database, similar to what Active Record provides. I was even thinking of mentioning it in the migration guide, but I didn't see official guides mentioning any 3rd-party gems, so I assumed it was a convention.

### DIFF
--- a/www/pages/plugins.html.erb
+++ b/www/pages/plugins.html.erb
@@ -1050,6 +1050,10 @@
 <span class="ul__span">A Sequel extension to make seeds/fixtures manageable like migrations.</span>
 </li>
 <li class="ul__li ul__li--grid">
+<a class="a" href="https://github.com/rosenfeld/sequel_tools">sequel_tools </a>
+<span class="ul__span">Tools to help with managing database operations with Sequel through Rake tasks.</span>
+</li>
+<li class="ul__li ul__li--grid">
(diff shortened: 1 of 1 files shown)


janko:
I considered this, but I saw that some gems listed in this section are technically not Sequel extensions either, such as fixture_dependencies and sequel-annotate. This one can be considered an extension of Sequel, just in the area of Rake tasks. If not here, where do you think these libraries should be linked?

AlexWayfer:
> I considered this, but I saw that some gems listed in this section are technically not Sequel extensions either, such as fixture_dependencies and sequel-annotate. This one can be considered an extension of Sequel, just in the area of Rake tasks. If not here, where do you think these libraries should be linked?

Have not seem them. Then… @jeremyevans decides.

But I'd like to see them separately. Like, "Tools" section or something like this.

I'm not against such projects and their mention, I just want to… kind of order them, sort, divide. Probably I'm wrong.

jeremyevans:
@janko Thanks for the patch!

@AlexWayfer I agree, "External Extensions" sort of implies things listed under the heading should be loadable by `Sequel.extension`.  However, since there isn't currently a distinction between actual extensions and just libraries that extend Sequel, I plan to merge this.  I'll happily accept another PR that splits the "External Extensions" section into separate sections, one for actual extensions and one for other libraries that extend Sequel.

PR #1955, opened 2022-10-25
Recommend disabling association reflection cache when using code reloading

janko:
*Follow-up to https://github.com/jeremyevans/sequel/pull/1951*

When using a code reloader that only reloads model classes that changed, the association reflection cache should be disabled, to avoid stale model classes still being referenced after they were reloaded.

This is not needed when using code reloaders like Zeitwerk, which always reload all the code regardless of which file changed, which includes all model classes. It's still recommended if you're reloading models after running schema migrations, as it causes the table schema cache to be automatically reloaded when the model is reloaded.

### DIFF
--- a/doc/association_basics.rdoc
+++ b/doc/association_basics.rdoc
@@ -342,6 +342,12 @@ instance method:
   @album.artists # [<Artist ...>, ...]
   @album.associations[:artists] # [<Artist ...>, ...]
 
+=== Code Reloading
+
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch!

jeremyevans:
Cherry-picked as 53b045879040128421f710d9a519318a849e92c7

PR #1962, opened 2022-11-13
Allow tactical eager loading in combination with class table inheritance

rolftimmermans:
Class table inheritance doesn't work in combination with tactical eager loading. There are no errors, but eager loading just isn't happening. I know this is intentional, but it makes it very difficult to manage performance when rendering a list of objects that use CTI if they make use of any association of the child classes.

This is a draft fix which seems to work nicely with our application. I'm opening this pull request to get your opinion. Is this on the right track? If so we'd be happy to polish and/or add more test cases.

The idea is that the `eager_load` method uses a concrete subclass rather than the parent model when performing eager loading. The tactical eager loading plugin takes care of passing in only the objects that are instances of the model on which the association was first used. This would always be a model that has access to this association.

I guess in theory it could still lead to some mild inefficiencies with very large hierarchies of inheritance. But that seems better than the status quo.

### DIFF
--- a/lib/sequel/model/associations.rb
+++ b/lib/sequel/model/associations.rb
@@ -3552,11 +3552,11 @@ module Sequel
         end
 
         # Eagerly load all specified associations.
-        def eager_load(a, eager_assoc=@opts[:eager])
+        def eager_load(a, m=model, eager_assoc=@opts[:eager])
           return if a.empty?
 
           # Reflections for all associations to eager load
-          reflections = eager_assoc.keys.map{|assoc| model.association_reflection(assoc) || (raise Sequel::UndefinedAssociation, "Model: #{self}, Association: #{assoc}")}
+          reflections = eager_assoc.keys.map{|assoc| m.association_reflection(assoc) || (raise Sequel::UndefinedAssociation, "Model: #{self}, Association: #{assoc}")}
 
           perform_eager_loads(prepare_eager_load(a, reflections, eager_assoc))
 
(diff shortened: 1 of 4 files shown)


lib/sequel/model/associations.rb
```diff
@@ -3552,11 +3552,11 @@ def eager_graph_loader
         end
 
         # Eagerly load all specified associations.
-        def eager_load(a, eager_assoc=@opts[:eager])
+        def eager_load(a, m=model, eager_assoc=@opts[:eager])
```
jeremyevans, line 3555: When adding an optional argument to an existing method, the optional argument should always be added at the end.

jeremyevans (changes requested):
Thanks for working on this.  I like this approach, it doesn't require `tactical_eager_loading` knowing anything about `class_table_inheritance`.  There is extra code when doing a tactical eager load that is unnecessary if `class_table_inheritance` is not used, but I can handle that post-merging.

Please see one inline comment.

rolftimmermans:
Great, thanks! The argument order has been swapped.

jeremyevans:
Thanks!  I'll try to get this merged tomorrow.

jeremyevans:
Squash merged as 23189e9ff3d382781a5ed18194266d5be1fec1cd

jeremyevans:
@rolftimmermans I made some changes to this support in 9a9f1119dfda85b10270763e5b0801b01cd401a7, can you review and let me know if that works for you?

PR #1969, opened 2022-12-01
Fix syntax in async_thread_pool extension docs

janko:
This fixes the syntax highlighting for these code blocks.


jeremyevans:
Thanks for the patch!

PR #1975, opened 2022-12-17
Calculate and validate min/max values for decimal types

rolftimmermans:
This patch detects the minimum and maximum values of fixed precision decimal types in order to allow automatic validation with the `auto_validations` plugin, similar to the current behaviour for integer types.

I tested this with a PostgreSQL database. I'm not 100% if other database also reject invalid values (as opposed to just truncating/rounding them). But I guess that if users turn on auto validations they values to be rejected if they don't fit the specified range?

### DIFF
--- a/lib/sequel/database/query.rb
+++ b/lib/sequel/database/query.rb
@@ -175,8 +175,13 @@ module Sequel
         if !c[:max_length] && c[:type] == :string && (max_length = column_schema_max_length(c[:db_type]))
           c[:max_length] = max_length
         end
-        if !c[:max_value] && !c[:min_value] && c[:type] == :integer && (min_max = column_schema_integer_min_max_values(c[:db_type]))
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for working on this.  From a brief review, this looks good.  I would definitely like to support this on more databases than just PostgreSQL.  I'll do some testing with this next week, and hopefully I can get this merged before the next release.

rolftimmermans:
Sounds good! 

While implementing I took types like `decimal(8)` into account (i.e. without any decimals). But currently Sequel treats them as integers. This means they also currently get integer min/max values, which is not strictly correct. I question the usefulness of such types, honestly, but at least now you know… 🙂

jeremyevans:
Cherry-picked at ad4da90595fefd274a46731fe5a86a207b3f3f66

rolftimmermans:
Awesome! 👏

PR #1983, opened 2023-01-18
Support enum value as array

keymastervn:
The existing enum plugin uses primitive values such as string or integer.
This PR enhances it by supporting array type to the enum values.

Background: on a large dataset of table X, I am working on schema migration, then data migration for column Y
Schema migration can be done easily, but data migration is not that quick. Y is currently nil.

To support multi values enum for my half-migrated case, I need to define the enum value as an array, obviously [1, nil] for column Y. Surprisingly, I can use an array for the value, the dataset method for query `:enum` work well
>#<Sequel::Postgres::Dataset: "SELECT * FROM \"X\" WHERE (\"Y\" IN (1, NULL))">

But other methods eg. instance enum assignment could not work as expected, that's why I make this PR.

### DIFF
--- a/lib/sequel/plugins/enum.rb
+++ b/lib/sequel/plugins/enum.rb
@@ -34,6 +34,14 @@ module Sequel
     # Album.dataset.bad:: Return a dataset filtered to rows where +status_id+ is +2+
     # Album.dataset.not_bad:: Return a dataset filtered to rows where +status_id+ is not +2+
     #
+    # The enum value can also be an array, for example:
+    #
+    #   Album.enum :status_id, good: [1, 2], bad: 3
+    #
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for sending a pull request. I don't think this is a common enough need to include this support in the enum plugin.  From looking at the implementation, it should be possible to write a plugin that depends on the enum plugin, overrides the methods and calls super to get the expected behavior.  You could even optimize it so that you only override methods when you know the enum value is an array.  Can you try creating this as an external plugin that depends on the enum plugin?

keymastervn:
Thank you @jeremyevans, let me create a plugin instead.

PR #1990, opened 2023-02-02
fix empty? for datasets containing null

andy-k:
`#empty?` was not returning `false` for some non-empty datasets.

```ruby
irb(main):001:0> require "sequel"
=> true
irb(main):002:0> DB = Sequel.sqlite
=> #<Sequel::SQLite::Database: {:adapter=>:sqlite}>
irb(main):003:0> DB["select null"].empty?
=> true
irb(main):004:0> DB["values (null, 1), (2, 3)"].empty?
=> true
irb(main):005:0> DB.values([[nil, 1], [2, 3]]).empty?
=> true
irb(main):006:0> 
```

### DIFF
--- a/lib/sequel/dataset/actions.rb
+++ b/lib/sequel/dataset/actions.rb
@@ -173,9 +173,9 @@ module Sequel
     #   DB[:table].empty? # SELECT 1 AS one FROM table LIMIT 1
     #   # => false
     def empty?
-      cached_dataset(:_empty_ds) do
+      with_sql_first((cached_dataset(:_empty_ds) do
         single_value_ds.unordered.select(EMPTY_SELECT)
-      end.single_value!.nil?
+      end).select_sql).nil?
     end
 
     # Returns the first matching record if no arguments are given.
--- a/spec/core/dataset_spec.rb
+++ b/spec/core/dataset_spec.rb
@@ -2279,6 +2279,14 @@ describe "Dataset#empty?" do
     db.sqls.must_equal ["SELECT 1 AS one FROM test WHERE 'f' LIMIT 1"]
   end
 
+  it "should return false if dataset contains record with null field" do
+    db = Sequel.mock(:fetch=>proc{|sql| {1=>nil} unless sql =~ /WHERE 'f'/})
+    db.from(:test).wont_be :empty?
+    db.sqls.must_equal ['SELECT 1 AS one FROM test LIMIT 1']
+    db.from(:test).filter(false).must_be :empty?
+    db.sqls.must_equal ["SELECT 1 AS one FROM test WHERE 'f' LIMIT 1"]
+  end
+
   it "should ignore order" do
     db = Sequel.mock(:fetch=>proc{|sql| {1=>1}})
     db.from(:test).wont_be :empty?


lib/sequel/dataset/actions.rb
```diff
@@ -173,9 +173,9 @@ def each
     #   DB[:table].empty? # SELECT 1 AS one FROM table LIMIT 1
     #   # => false
     def empty?
-      cached_dataset(:_empty_ds) do
+      with_sql_first((cached_dataset(:_empty_ds) do
```
jeremyevans, line 176: We should leave the existing code as is for datasets not using custom SQL, because it prevents loading a large amount of data.  For datasets using custom SQL, maybe we should call `from_self`, either in this method, or in `single_value_ds` if this also affects other methods.  If in this method, maybe something like:

```diff
-        single_value_ds.unordered.select(EMPTY_SELECT)
+        (@opts[:sql] ? from_self : self).single_value_ds.unordered.select(EMPTY_SELECT)
```

jeremyevans (changes requested):
Thanks for finding and fixing this issue for datasets using custom SQL.  Please see requested changes regarding the implementation.

spec/core/dataset_spec.rb
```diff
@@ -2279,6 +2279,14 @@ def ==(o) @h == o.h; end
     db.sqls.must_equal ["SELECT 1 AS one FROM test WHERE 'f' LIMIT 1"]
   end
 
+  it "should return false if dataset contains record with null field" do
```
jeremyevans, line 2282: The issue you are trying to fix is only present for datasets using custom SQL (no SQL server will return a NULL value for SELECT 1).  So the spec should also be modified to use custom SQL.

lib/sequel/dataset/actions.rb
andy-k: Thanks for the suggestion. I tried it (making the suggested change and reverting the lines above and below it), it fixes the custom SQL cases, but makes `DB.values([[nil, 1], [2, 3]]).empty?` incorrect again as it is not custom SQL. How can we also detect this case, and are there any other cases that need to be worked around?

My thought process when I came up with the code I shared was as follows, please correct my assumptions in the comments:
```ruby
# original code for reference, unchanged
def with_sql_single_value(sql)
  if r = with_sql_first(sql)
    r.each{|_, v| return v}
  end
end

# original code for reference, unchanged
def single_value!
  with_sql_single_value(select_sql)
end

# 1. original code,
# where dataset = (cached_dataset(...) {...}) since that is unchanged
def empty?
  dataset.single_value!.nil?
end

# 2. inline the call to single_value!
def empty?
  with_sql_single_value(dataset.select_sql).nil?
end

# 3. inline the call to with_sql_single_value
def empty?
  (if r = with_sql_first(dataset.select_sql)
    r.each{|_, v| break v}
  end).nil?
end

# 4. fix the bug: if there is a row we should return false,
# without checking if the first iterated column of the row is also not nil
def empty?
  (if r = with_sql_first(dataset.select_sql)
    r
  end).nil?
end

# 5. simplifying, assuming with_sql_first never returns false
def empty?
  with_sql_first(dataset.select_sql).nil?
end
```

Would you mind sharing how the code prevents loading a large amount of data? I'm not yet familiar enough with the library and was unaware of this case and how to trigger it. Moreover, I'm not sure which part of the above transformation would break this (loading a large amount of data) case specifically.

spec/core/dataset_spec.rb
andy-k: I'll need help with modifying the spec to use custom SQL.

Also, we will need tests for `DB.values([[nil, 1], [2, 3]]).empty?`, any other non-custom SQL cases that also need to be worked around, and maybe the loading a large amount of data case. I'll need help with these.

andy-k (commented):
Thank you for the review! Please see comments regarding the requested changes.

jeremyevans:
> Would you mind sharing how the code prevents loading a large amount of data? I'm not yet familiar enough with the library and was unaware of this case and how to trigger it. Moreover, I'm not sure which part of the above transformation would break this (loading a large amount of data) case specifically.

Consider the following:

```ruby
DB.create_table(:t){String :s, :text=>true}
DB[:t].insert('1'*100_000_000)

DB[:t].empty?
# SELECT 1 AS 'one' FROM `t` LIMIT 1

DB['SELECT * FROM t'].empty?
# Before and with your change:
# SELECT * FROM t
# With my recommended change:
# SELECT 1 AS 'one' FROM (SELECT * FROM t) AS 't1' LIMIT 1
```

Your change fixes the return value (good), but it will load a large value (bad).  My approach fixes the return value, but avoids loading a large value.  You can benchmark the above cases, but my approach is probably at least 50x faster in the above example, compared to yours.

The `Dataset#values` case is different.  For both SQLite and PostgreSQL, the `VALUES` statement requires at least one row, so `Dataset#empty?` could always return true without a query in that case.  We should probably make `Dataset#values` raise for empty values.

I can build on top of this pull request and fix the remaining issues.  This will need integration tests and testing on all databases Sequel supports to ensure correct behavior.

andy-k:
I see. `DB['SELECT * from t'].empty?` loads the large value in the "with my change" case. I misunderstood your comment as that I had introduced a regression, and am now relieved that I hadn't; this is exactly the same behavior as the "before" case. In this PR, I was primarily interested in fixing the return value of `#empty?`.

`select 1 from ...` would definitely help performance in the specific case when the original query selects a big value and that big value happens to be the first row returned. But that's rare, and there is a minimal cost to build that wrapper; there is a reason we aren't just wrapping everything with `#from_self` regardless whether it's a custom SQL, a `#values`, or just `DB[:t]`. (With PostgreSQL (but not SQLite), you can even `select from t` and get rows of nothings.)

I disagree to have all custom SQL automatically wrapped in `#from_self`. It is easy enough for the caller to do `DB['SELECT * from t'].from_self.empty?` manually when they believe this is a real optimization that applies in their case; in most cases, the query looks more like `DB['SELECT * from t order by something'].empty?` and that's not being helped by `#from_self`. When the caller already constructs an `.empty?`-friendly query -- in my case I constructed `select null from ...` and found this issue -- it is probably not beneficial to (incur GC and) forcibly convert that to `select 1 from (select null from ...)`. I might be in the minority, I had assumed `def empty? = !first` all along and actually expected the whole first row to be loaded... 😅 

Yes, feel free to build on top of this pull request, optimize `Dataset#values`, and so on.

Optimizations and different approaches aside, would you agree that this pull request addresses the core issue (the return value of `#empty?`) without any regressions?

Thanks for the library!

jeremyevans:
> `select 1 from ...` would definitely help performance in the specific case when the original query selects a big value and that big value happens to be the first row returned. But that's rare, and there is a minimal cost to build that wrapper; there is a reason we aren't just wrapping everything with `#from_self` regardless whether it's a custom SQL, a `#values`, or just `DB[:t]`. (With PostgreSQL (but not SQLite), you can even `select from t` and get rows of nothings.)

The wrapper is cached, so it's only built the first time it is needed for the dataset.  It's true that we aren't using `from_self` in all cases.  There isn't a reason to use it if you know you don't need it.

> I disagree to have all custom SQL automatically wrapped in `#from_self`. It is easy enough for the caller to do `DB['SELECT * from t'].from_self.empty?` manually when they believe this is a real optimization that applies in their case; in most cases, the query looks more like `DB['SELECT * from t order by something'].empty?` and that's not being helped by `#from_self`.

Without the `from_self`, if any column in `t` is large, this can significantly slow down `empty?`.  Note that there are lots of other cases in Sequel where we wrap datasets with custom SQL in `from_self`, such as:

```ruby
DB['SELECT * FROM t'].count
# SELECT count(*) AS "count" FROM (SELECT * FROM t) AS "t1" LIMIT 1
```

So following this practice with `empty?` is consistent with Sequel's behavior in other cases, and a good idea for the reasons I've discussed.

> When the caller already constructs an `.empty?`-friendly query -- in my case I constructed `select null from ...` and found this issue -- it is probably not beneficial to (incur GC and) forcibly convert that to `select 1 from (select null from ...)`. I might be in the minority, I had assumed `def empty? = !first` all along and actually expected the whole first row to be loaded... 😅

Asserting one approach is faster than another without benchmarking is a risky business.  Here's a much more reasonable example with a table with only 6 columns:

```ruby
DB.create_table(:t) do
  primary_key :id
  5.times do |i|
    String :"c#{i}"
  end
end
s = '1' * 50
DB[:t].insert(:c0=>s, :c1=>s, :c2=>s, :c3=>s, :c4=>s)

ds = DB['SELECT * FROM t']
require 'benchmark/ips'
Benchmark.ips do |x|
  x.report('empty?'){ds.empty?}
end
```

Results with your approach:

```
empty?     12.734k (_ 0.8%) i/s -     64.719k in   5.082693s
```

Results with my approach:

```
empty?     18.525k (_ 1.5%) i/s -     94.401k in   5.097011s
```

> Yes, feel free to build on top of this pull request, optimize `Dataset#values`, and so on.

OK.  I'll build on top of this and make sure this issue is fixed before the next release.

> Optimizations and different approaches aside, would you agree that this pull request addresses the core issue (the return value of `#empty?`) without any regressions?

Certainly your pull request fixes the issue for datasets with custom SQL and those using `values`, and I thank you for that.

andy-k:
Thanks for merging and for adjusting!

Unfortunately, it seems the commit that follows the merge (a638268ce08892cd06d7b858a2b9770f0ed2e183) broke this.
```ruby
# this PR #1990 made DB.values(...).empty? return false
$ git checkout a638268ce^
HEAD is now at 23534e85d fix empty? for datasets containing null
$ ruby -r './lib/sequel' -e 'p Sequel.sqlite.values([[nil]]).empty?'
false

# the commit following it made it return true again
$ git checkout a638268ce
Previous HEAD position was 23534e85d fix empty? for datasets containing null
HEAD is now at a638268ce Avoid loading full row in Dataset#empty?
$ ruby -r './lib/sequel' -e 'p Sequel.sqlite.values([[nil]]).empty?'
true

# and that's the same as how it was before
$ git checkout a638268ce^^
Previous HEAD position was a638268ce Avoid loading full row in Dataset#empty?
HEAD is now at a534d716d Bump version to 5.65.0
$ ruby -r './lib/sequel' -e 'p Sequel.sqlite.values([[nil]]).empty?'
true
```

It seems that [the `return false` added in the commit](https://github.com/jeremyevans/sequel/commit/a638268ce08892cd06d7b858a2b9770f0ed2e183#diff-9a07e9ea825021ea7117238d04accfc79dc951839fae70b8c8b9233b03e60c01R661) is not executed...

jeremyevans:
I think your issue is you are only requiring the Sequel library without changing the load path, so you are not loading the sqlite adapter for the checkout of Sequel, your are probably loading the sqlite adapter from an installed sequel gem.

There was a spec added for the behavior, and you can also see it works correctly on the command line:

```
$ bin/sequel sqlite:/ -c 'p DB.values([[nil]]).empty?'    
false
```

andy-k:
You're right, there's no regression. Thanks for patiently sharing the correct command line to use for testing this!

PR #1998, opened 2023-02-17
alias key? has_key?

TSMMark:
`Hash#key?` and `Hash#has_key?` are aliases of each other in Ruby. This would bring pg_json_ops in line with similar aliases.

Additionally, Rubocop has Style/PreferredHashMethods which prefers `key?` over `has_key?` by default. Of course rubocop does not know which objects are Hashes and which are pg json op, it registers offenses when using `has_key?` on pg json op objects.

This would be a small QOL improvement for users of rubocop, and all accustomed to `key?` over `has_key?`

### DIFF
--- a/lib/sequel/extensions/pg_json_ops.rb
+++ b/lib/sequel/extensions/pg_json_ops.rb
@@ -421,6 +421,7 @@ module Sequel
         bool_op(HAS_KEY, key)
       end
       alias include? has_key?
+      alias key? has_key?
 
       # Inserts the given jsonb value at the given path in the receiver.
       # The default is to insert the value before the given path, but


benkoshy:
> All that being said, I'll consider this if there is broad community support for it. So I'll leave this open for a while to collect feedback from the community.

I appreciate the PR - I thank you for providing it. I also appreciate the maintainer's view - which FWIW, I concur with.

(I use sequel in a hobby context - small personal protects - so perhaps weigh this opinion accordingly).

jeremyevans:
I'm closing this now.  It's easy to add the alias itself if you need it.

tycooon:
Actually there is a bunch of issues that people might have when using Rubocop with Sequel. For example, we have the following Rubocop configs in one of our projects:

```yaml
Style/NumericPredicate:
  Enabled: false # Doesn't work with sequel

Style/PreferredHashMethods:
  Enabled: false # Doesn't work with sequel

Style/ConcatArrayLiterals:
  Enabled: false # Doesn't work with sequel
```

 So maybe at some point someone will come up with a plugin that adds a bunch of "missing" aliases like the one in this PR.

jeremyevans:
> So maybe at some point someone will come up with a plugin that adds a bunch of "missing" aliases like the one in this PR.

I'd be happy to add a link on Sequel's website to an external extension that did that.

PR #2003, opened 2023-03-09
Add bigdecimal as an explicit dependency

ekohl:
While sequel always requires bigdecimal, it's not listed as a dependency. It's possible to have a Ruby installation without bigdecimal (on EL8 / Fedora: `dnf install ruby --setopt=install_weak_deps=False`) and this helps. While Fedora already explicitly adds it as a dependency, listing it in the gemspec allows the dependency generator to pick it up automatically.

### DIFF
--- a/sequel.gemspec
+++ b/sequel.gemspec
@@ -23,6 +23,7 @@ SEQUEL_GEMSPEC = Gem::Specification.new do |s|
   s.require_path = "lib"
   s.bindir = 'bin'
   s.executables << 'sequel'
+  s.add_dependency "bigdecimal"
   s.add_development_dependency "minitest", '>=5.7.0'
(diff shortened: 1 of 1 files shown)


ekohl:
After I opened a similar PR in Rails it was pointed out bigdecimal is actually a candidate for moving from a default gem to a bundled gem in Ruby 3.3: https://bugs.ruby-lang.org/issues/19351.

AlexWayfer:
If you use installation method like "don't install deps" — it's a problem of installation method, yes, I agree.

In other hand, if we already have the official `bigdecimal` gem, and we can lock it better (major better versions), like with `psych` right now — I'd like to add this as dependency. I see nothing wrong with it, only pros.

PR #2009, opened 2023-03-26
chore: add new issues template

benkoshy:
### What is this?

* add issues templates.

### Why do i need this?

* to protect maintainers' time. issues are still reported as bugs, when they really should be questions:
* A streamlined user interface to encourage posting on the dicussion forum may help: the same has been implemented on the shrine gem and I don't think i've seen a single mis-allocated issue.

### Demo

Try it out here on my repo: https://github.com/benkoshy/sequel/issues/new/choose

image example:

options_to_choose


jeremyevans:
Thanks for the patch! This seems like an improvement.  I'll merge this and tweak it.

benkoshy:
thx for the gem and also the maintenance.

i'll throw in my 2 cents: A link to a fully self-contained sample template may help:

- it eases the conceptual complexity (especially for new comers) to put one together in the first place (I wouldn't even know where to start). Some might discover the source of the "error" in doing so.
- reduces maintenance burden: very difficult to fix a bug without being able to reproduce it.
- It gives opportunity for bystanders to triage issues. 

add_sample_template

Shrine hits it out of the park: https://github.com/shrinerb/shrine/blob/master/SELF_CONTAINED_EXAMPLE.md#option-2-manually-set-up-the-self-contained-example

shrine

jeremyevans:
> i'll throw in my 2 cents: A link to a fully self-contained sample template may help:
> 
> * it eases the conceptual complexity (especially for new comers) to put one together in the first place (I wouldn't even know where to start). Some might discover the source of the "error" in doing so.

Almost every bug is different.  I'm not sure what a sample template would look like.  I'm open to suggestions, though.  However, I'm 100% against recommending someone clone a repository to report a bug.  Ideally, the self contained example will be a few lines of Ruby code.  If someone links to a repository that reproduces the issue, the first thing I am going to ask them to do is turn it into a single Ruby file and post the code for it (i.e. create the simplest possible example).

> * reduces maintenance burden: very difficult to fix a bug without being able to reproduce it.

That's why the existing template asks for the simplest possible self contained example.

> * It gives opportunity for bystanders to triage issues.

This should be true regardless of whether there is a sample template or not.

benkoshy:
I appreciate the thoughtful consideration.

PR #2011, opened 2023-03-28
Fix small typo in documentation

AlexWayfer:
I'm pretty sure there should be the underscore.

I haven't found other similar places in this file.


jeremyevans:
You are correct.  Thanks for the patch!

PR #2017, opened 2023-04-11
Replace gsub with tr on_delete_clause method

RicardoTrindade:
As per https://github.com/fastruby/fast-ruby/blob/main/code/string/gsub-vs-tr.rb `tr` is faster than `gsub` for simple replacements that do not require regexp


jeremyevans:
Thanks for the patch!  This isn't in performance sensitive code, but as `tr` is simpler than `gsub`, it's still a good change to make.

PR #2036, opened 2023-05-04
Fix referencing Trilogy error class when disconnecting

janko:
I was looking at this commit in order to add support for the Trilogy adapter in sequel-activerecord_connection, when I noticed this and thought I'd send a pull request. It's very nice this will be supported :metal:


jeremyevans:
Thanks for the patch!

PR #2040, opened 2023-05-17
Expose the model class inside the dataset module

davekaro:
One reason you might want this is to qualify the columns in your method. For example, `where :kept, Sequel[model.table_name][:discarded_at] => nil`

See discussion https://github.com/jeremyevans/sequel/discussions/2038. I'm not sure the test is up to standards, I struggled at first so happy for any feedback/adjustments needed.


davekaro:
Sorry didn't notice the whitespace changes... will follow up shortly.

jeremyevans:

PR #2043, opened 2023-05-26
Sharding docs: Cleary state that :read_only is a special key

TSMMark:
One of my senior teammates is not convinced from the documentation that `:read_only` is a special key that induces a certain automatic behavior with how read vs write queries are handled.

After I read the docs again, I was not sure anymore either. It seems implied, but never clearly stated.

This PR adds additional description that is to my best understanding. 

Please update or correct as needed. I feel that some additional clarification here could definitely help Sequel users


jeremyevans:
Thanks for the patch!  I'll merge and reword.

PR #2048, opened 2023-06-08
Prevent connection pool starvation.

ioquatix:
See <https://github.com/socketry/async/issues/99> for context/background.

This should be a temporary fix, we need to (1) establish whether there is a bug in CRuby or (2) establish whether there is faulty usage of condition variable in the connection pool. I don't know which yet.

The code related to this investigation is <https://github.com/ioquatix/ruby-condition-variable-timeout>.


jeremyevans:
Thanks for the patch!  Running tests with it now.  Assuming no fallout, I'll merge this.

Issue #2042, opened 2023-05-25
connection_validator plugin: "undefined method `sync'" bug on single threaded connection


HoneyryderChuck:
### Complete Description of Issue

When loading it on a database object marked as single threaded, a `NoMethodError` is raised.

### Simplest Possible Self-Contained Example Showing the Bug

```ruby
DB = Sequel.connect(ENV["DATABASE_URL"], single_threaded: true)
DB.extension(:connection_validator) #=> undefined method `sync' for #<Sequel::SingleConnectionPool:0x00007fedc6959d98 @db
# the second call works, possibly because the original object has been marked as disconnected?
DB.extension(:connection_validator) #=> #<Sequel::Postgres::Database: "...
```

### Full Backtrace of Exception (if any)

/usr/local/bundle/ruby/3.2.0/gems/sequel-5.68.0/lib/sequel/extensions/connection_validator.rb:65:in `block in extended': undefined method `sync' for #<Sequel::SingleConnectionPool:0x00007fedc6959d98 @db=#<Sequel::Postgres::Database: "postgresql://postgres:password@database/postgres" {:single_threaded=>true}>, @use_old_connect_api=false, @after_connect=nil, @connect_sqls=nil, @error_classes=[PG::Error], @conn=[#<Sequel::Postgres::Adapter:0x00007fedc6942f30 host=database port=5432 user=postgres type_map_for_queries=#<PG::TypeMapByClass:0x00007fedc6cd2a10>>]> (NoMethodError)

        sync do
        ^^^^
        from /usr/local/bundle/ruby/3.2.0/gems/sequel-5.68.0/lib/sequel/extensions/connection_validator.rb:64:in `instance_exec'
        from /usr/local/bundle/ruby/3.2.0/gems/sequel-5.68.0/lib/sequel/extensions/connection_validator.rb:64:in `extended'
        from /usr/local/bundle/ruby/3.2.0/gems/sequel-5.68.0/lib/sequel/extensions/connection_validator.rb:121:in `extend'
        from /usr/local/bundle/ruby/3.2.0/gems/sequel-5.68.0/lib/sequel/extensions/connection_validator.rb:121:in `block in <module:Sequel>'
        from /usr/local/bundle/ruby/3.2.0/gems/sequel-5.68.0/lib/sequel/database/misc.rb:236:in `block in extension'
        from /usr/local/bundle/ruby/3.2.0/gems/sequel-5.68.0/lib/sequel/database/misc.rb:233:in `each'
[...]

PR #2049, opened 2023-06-08
raise error when connection_validator is loaded in single threaded mode

HoneyryderChuck:
While, as per the discussion in #2042 , `sequel` is doing what is supposed to as per what's documentedd, the error is still cryptic and not obvious when one falls in the trap.

This patch adds a more helpful error message, which is aligned with other similar messages for other extensions which aren't supported in single threaded mode as well.

### DIFF
--- a/lib/sequel/extensions/connection_validator.rb
+++ b/lib/sequel/extensions/connection_validator.rb
@@ -61,6 +61,10 @@ module Sequel
 
     # Initialize the data structures used by this extension.
     def self.extended(pool)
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch.  I'll do some testing with it tomorrow.  It looks like the connection_validator extension is not compatible with the timed_queue connection pool, which should definitely be fixed.

jeremyevans:
Cherry picked as acb8bca5ebd7d61a9765835eab1ddfea37999658

PR #2060, opened 2023-07-03
Make set_column_allow_null reversible

enescakir:
Reverse of "set_column_allow_null" operation is calling it with inverse of "allow_null" argument.


jeremyevans:
This looks good.  Thanks for the patch!

PR #2066, opened 2023-07-27
support ILIKE ANY in Postgres, similar to LIKE ANY

gilesbowkett:
this PR applies the same change from 9e9a5d7eadfba6f549eb7adbd6000a4fbf6ff3fc to Postgres-specific `ILIKE` code, with the same intention: to support use cases like `ILIKE ANY`.

since the feature method is hard-coded to `true`, checking its value is not actually necessary, but implementing the change this way means that this code looks just like the other code in `lib/sequel/dataset/sql.rb` which it is emulating.

that code is here:

https://github.com/jeremyevans/sequel/blob/c280ff50517e7a8dccc5e3a56ba2e316c4379b91/lib/sequel/dataset/sql.rb#L395

apologies for the lack of tests. the contributing guidelines say they're needed for any "substantial bug fixes," and this is just a two-line fix. but I can take a second to write some tests up, if needed.

thanks!

### DIFF
--- a/lib/sequel/adapters/shared/postgres.rb
+++ b/lib/sequel/adapters/shared/postgres.rb
@@ -1745,10 +1745,6 @@ module Sequel
           literal_append(sql, args[0])
           sql << ' ' << op.to_s << ' '
           literal_append(sql, args[1])
-          if requires_like_escape?
-            sql << " ESCAPE "
-            literal_append(sql, "\\")
-          end
           sql << ')'
         else
           super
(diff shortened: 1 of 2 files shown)


jeremyevans:
I'm fine with the general idea.  Seems like you could just delete the lines, since `requires_like_escape?` is false on PostgreSQL and this is PostgreSQL specific code.  I'll try make sure this makes the next release.

jeremyevans:
Also, thank you very much for submitting a patch to fix this, I appreciate it!

gilesbowkett:
ok, it somehow took me four commits because it's late at night, but I deleted the extra lines.

anyway, happy to help!

jeremyevans:
Squash merged at b047c73df3fb7b0cd6314702775c53578190d01b

PR #2068, opened 2023-08-03
Fix spelling utrused->untrusted

mikepmunroe:


jeremyevans:
Thanks for the patch!

PR #2071, opened 2023-08-15
Fix documentation typo

amatsuda:
assocation => association


jeremyevans:
Thanks for the patch!

PR #2079, opened 2023-09-14
Pass class name to Sequel::MassAssignmentRestriction error

artofhuman:
If the instance has a class name, pass this name to the error message to better understand the error in monitoring systems such as Sentry.

### DIFF
--- a/lib/sequel/model/exceptions.rb
+++ b/lib/sequel/model/exceptions.rb
@@ -31,9 +31,10 @@ module Sequel
     attr_reader :model
 
     def self.create(msg, model)
-      @model = model
       msg += " for class #{model.class.name}" if model.class.name
-      new(msg)
+      new(msg).tap do |err|
+        err.instance_variable_set(:@model, model)
+      end
     end
   end
   
--- a/lib/sequel/model/exceptions.rb
+++ b/lib/sequel/model/exceptions.rb
@@ -24,11 +24,18 @@ module Sequel
   UndefinedAssociation = Class.new(Error)
   ).name
 
-  (
   # Raised when a mass assignment method is called in strict mode with either a restricted column
   # or a column without a setter method.
-  MassAssignmentRestriction = Class.new(Error)
-  ).name
+  class MassAssignmentRestriction < Error
+    # The Sequel::Model object related to this exception.
+    attr_reader :model
+
+    def initialize(msg, model)
+      @model = model
+      msg += " for class #{model.class.name}" if model.class.name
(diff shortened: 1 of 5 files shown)


jeremyevans:
Instead of including the model class name in the message, we should add a method to set the model instance as an instance variable in the exception.  That takes less work, and still allows the rescuer to get the class name.  This should be implemented using a class method on `MassAssignmentRestriction`, which the callers can use.

artofhuman:
Added `@model` instance variable to `MassAssignmentRestriction`, similar to `ValidationFailed`

lib/sequel/model/exceptions.rb
```diff
@@ -24,11 +24,18 @@ def initialize(message=nil, model=nil)
...
-  (
   # Raised when a mass assignment method is called in strict mode with either a restricted column
   # or a column without a setter method.
-  MassAssignmentRestriction = Class.new(Error)
-  ).name
+  class MassAssignmentRestriction < Error
+    # The Sequel::Model object related to this exception.
+    attr_reader :model
+
+    def initialize(msg, model)
```
jeremyevans, line 33: This is not backwards compatible, and will break code that uses `raise MassAssignmentRestriction`.  I recommend using a different class method, such as `create`.

artofhuman: Maybe better to make model arg optional?

jeremyevans: I would prefer a separate method.

lib/sequel/model/exceptions.rb
```diff
@@ -24,11 +24,18 @@ def initialize(message=nil, model=nil)
...
   # Raised when a mass assignment method is called in strict mode with either a restricted column
   # or a column without a setter method.
-  MassAssignmentRestriction = Class.new(Error)
-  ).name
+  class MassAssignmentRestriction < Error
+    # The Sequel::Model object related to this exception.
+    attr_reader :model
+
+    def self.create(msg, model)
+      @model = model
```
jeremyevans, line 34: This sets the instance variable on the class, not on the created instance.  You don't catch this because you only test for the message and not for the model value.  The tests should be updated to test that the model is set correctly.

jeremyevans:
Thanks for continuing to work on this.  I'll try to merge and clean this up on Monday.

jeremyevans:
Squash merged at 1e00c84d23877b1a6395b76c761b60ded9dcf5a7

PR #2081, opened 2023-09-23
Rename paged_update_delete plugin to paged_operations

janko:
As proposed in https://github.com/jeremyevans/sequel/discussions/2080, now the paged_update_delete plugin supports `#paged_datasets` that can be used for other operations, we rename it to paged_operations to be more generic.


jeremyevans:
Thanks for the patch!  I'll test and merge shortly.

PR #2082, opened 2023-09-23
Minor fixes to paged_operations plugin docs

janko:
Update places that still refer to `paged_operations_size` to use `:rows_per_page`, and also replace `:rows_per_fetch` which is a `#paged_each` option.


jeremyevans:
Thanks for the patch!

PR #2093, opened 2023-11-09
Added run_single method to run single migration by version and direction

opya:
The new method was added only for TimestampMigrator because I'm not sure that this is the correct way to do it.

Reference: https://github.com/jeremyevans/sequel/discussions/2045

### DIFF
--- a/lib/sequel/extensions/migration.rb
+++ b/lib/sequel/extensions/migration.rb
@@ -426,10 +426,6 @@ module Sequel
       migrator_class(directory).new(db, directory, opts).run
     end
 
-    def self.run_single(db, path, opts=OPTS)
-      new(db, File.dirname(path), opts).run_single(path, opts[:direction])
-    end
-
     # Choose the Migrator subclass to use.  Uses the TimestampMigrator
     # if the version number is greater than 20000101, otherwise uses the IntegerMigrator.
     def self.migrator_class(directory)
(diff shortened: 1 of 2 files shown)


jeremyevans (changes requested):
I think the API should be different:

```ruby
  Sequel::TimestampMigrator.run_single(DB, '/path/to/filename', opts)
```

With this definition:

```ruby
  def TimestampMigrator.run_single(db, path, opts=OPTS)
    new(db, File.dirname(path), opts).run_single(path, opts[:direction])
  end
```

This avoids a lot of the unnecessary code in this pull request.  It also handles the case of multiple migrations with the same version (allowed by the TimestampMigrator). Can you switch to that approach?

opya:
sorry for the multiple commits, i think that the requested comments were addressed, sorry if forgot something

lib/sequel/extensions/migration.rb
```diff
@@ -585,6 +589,10 @@ def run
       target
     end
 
+    def run_single(path, direction)
```
jeremyevans, line 592: This code should be removed.

jeremyevans (changes requested):
Looking better.  Please see some additional requested changes.

jeremyevans (commented):
This looks good.  I'll try to merge and test on Monday.

opya:
ok, thank you

jeremyevans:
Squash merged at 422f16040a09e5bf5f6298121f5e246f91418ea4

PR #2096, opened 2023-11-12
Remove unused Rake tasks

v-kolesnikov:
These are unused since ce5b073d47

```
$ rake spec_w
rake aborted!
Don't know how to build task 'spec_core_w' (See the list of available tasks with `rake --tasks`)
Did you mean?  spec_core
               spec_core_ext
```


jeremyevans:

PR #2098, opened 2023-11-16
Add support for hostaddr param in postgres

mscrivo:
This is useful for when you want to tell your connection specifically which host IP to connect to but still optionally use hostname for authentication with tls.

Supported in pg gem since 1.4.0 and postgres since version 12. I'm not sure if we need to codify those constraints?

### DIFF
--- a/doc/opening_databases.rdoc
+++ b/doc/opening_databases.rdoc
@@ -348,7 +348,8 @@ The following additional options are supported:
                 <tt>Sequel.postgres(search_path: ['schema1', 'schema2'])</tt>).
 :use_iso_date_format :: This can be set to false to not force the ISO date format.  Sequel forces
                         it by default to allow for an optimization.
-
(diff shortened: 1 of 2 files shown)


jeremyevans:
I think this is supported already via `driver_options: {hostparam: '12.34.56.78'}`.  Can you try that and see if it works?  If so, I don't think we want to add another top-level specific option for it, as the need for it seems uncommon.

mscrivo:
> I think this is supported already via `driver_options: {hostparam: '12.34.56.78'}`. Can you try that and see if it works? If so, I don't think we want to add another top-level specific option for it, as the need for it seems uncommon.

ah I didn't realize that was a thing, I'll give that a shot and let you know. Thanks!

mscrivo:
`driver_options: {hostaddr: '12.34.56.78'}` did indeed work! Thanks @jeremyevans

PR #2099, opened 2023-11-16
Fix Ruby syntax in column encryption constraint example

dentarg:


jeremyevans:
Thanks for the patch!

PR #2107, opened 2023-12-19
Update rcte plugin to allow the user to specify union_all in rcte

JonathanFrias:
Support UNION style recursive common table expressions to remove duplicates by exposing a passthrough option :union_all

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 === master
 
+* Support generating rcte queries using UNION or UNION ALL in the rcte plugin (jonathanfrias)
+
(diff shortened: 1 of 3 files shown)


jeremyevans:
Can you explain why you want this?  I'm not opposed to this, but I'm not sure in which cases it would be helpful.  It doesn't seem like it would help lazy loading, since that should not result in duplicate rows in a tree.  Eager loading where the base results are in different levels of the tree maybe?

JonathanFrias:
I have a tree based data structure where there could be many duplicates in many levels in the tree, but for my use case I only care if there's 1, so I'm looking at the feasibility of using this plugin.

PR #2108, opened 2023-12-22
Improve performance of regexp matching in the sqlite adapter

paddor:
Nothing much, just what the title says. Theoretically, if an infinite amount of different regexp are used in SQL queries, it could of course fill up RAM, but I think that's uncommon.

In my case, this brought the runtime from ~300ms down to ~70ms for a query that uses one single regexp in a table with around ~5700 rows.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -24,7 +24,7 @@
 
 * Add auto_cast_date_and_time extension, for casting date and time values using SQL standard functions (jeremyevans)
 
-* Cache regexp objects in the sqlite adapter to improve performance (paddor)
+* Improve performance and flexibility of regexp matching in sqlite adapter (paddor)
 
 === 5.75.0 (2023-12-01)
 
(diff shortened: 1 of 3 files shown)


jeremyevans (changes requested):
I'm OK with supporting this, but it must be conditional and off by default for backwards compatibility.  This introduces a potential memory leak for code that does not expect the regexps to be cached.

paddor:
I've changed the implementation. It now only caches Regexp objects if `setup_regexp_function => :cached` was given.

paddor:
Thinking about it a little longer, maybe the option `setup_regexp_function` should just support taking a block.

paddor:
This is how I plan to use it now:

```ruby
regexp_cache = Hash.new { |h, k| h[k] = Regexp.new(k, timeout: 0.005) }
opts = {
  setup_regexp_function: proc { |re_str, str| regexp_cache[re_str].match? str }
}
Sequel.connect('sqlite:///path/db.sqlite3', opts)
```

Let me know what you think. I left the `:cached` functionality in, with appropriate changes to the documentation. Let me know if I should remove it.

jeremyevans:
Squash merged at c8248241ddf1a8d5d06f296aab5dca40b777c7b8

paddor:
Thanks!

PR #2115, opened 2024-01-05
Honor configured `:top` option for empty lists

johanmagnusson:
Using the list plugin with the option `:top` set to something other than the default (`1`) is causing discrepancies when adding a new element to an empty list without providing a specific position value. The plugin will always set the start value to `1` and ignore the configured top option.

This PR will make the list plugin honor the configured top option when the first item in the list is added when the position value is not provided.


jeremyevans:
Thanks for the patch! I agree this is a bug and the fix looks good.

PR #2122, opened 2024-01-19
Use empty? instead of count for validate uniqueness

numbata:
Using `empty` provides a more efficient way to check for records that violate the uniqueness constraint. Unlike `count`, which counts all matching rows, `empty` stops as soon as it finds a match, thereby reducing the load on the database. This change is particularly beneficial for tables with a large number of rows, where performance can be an issue.

### DIFF
--- a/spec/extensions/auto_validations_spec.rb
+++ b/spec/extensions/auto_validations_spec.rb
@@ -2,7 +2,15 @@ require_relative "spec_helper"
 
 describe "Sequel::Plugins::AutoValidations" do
   before do
-    db = Sequel.mock(:fetch=>proc{|sql| sql =~ /'a{51}'|'uniq'/ ? {:v=>0} : {:v=>1}})
+    fetch_proc = proc do |sql|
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch.  This makes sense to me.  I'll review and test later today.

spec/extensions/auto_validations_spec.rb
```diff
@@ -2,7 +2,15 @@
 
 describe "Sequel::Plugins::AutoValidations" do
   before do
-    db = Sequel.mock(:fetch=>proc{|sql| sql =~ /'a{51}'|'uniq'/ ? {:v=>0} : {:v=>1}})
+    fetch_proc = proc do |sql|
+      if sql =~ /'a{51}'|'uniq'/
+        sql =~ /1 AS one/ ? {:v=>nil} : {:v=>0}
```
numbata, line 7: This is not a style change. The code was unwrapped because new check was added here.

PR #2143, opened 2024-03-20
Declare csv dependency

ebarendt:
## Problem

[Ruby 3.3 warns](https://github.com/jeremyevans/sequel/actions/runs/8273014466/job/22635954106):

```
/home/runner/work/sequel/sequel/spec/extensions/csv_serializer_spec.rb:3: warning: csv was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add csv to your Gemfile or gemspec.
```

## Solution

Declare the CSV library as a dependency.


ebarendt:
I guess this is a plugin, unused otherwise. Anyone using csv probably has this declared already.

PR #2144, opened 2024-03-27
Make it possible to use create_or_replace_view with materialized view.

nashby:
It's not possible to use CREATE OR REPLACE with materialized view in Postgres so we should fallback to `drop_view`/`create_view` calls when we use `create_or_replace_view` method with `:materialized` option.

Currently when `create_or_replace_view` is called you get

```
Sequel::DatabaseError: PG::SyntaxError: ERROR:  syntax error at or near "MATERIALIZED"
LINE 1: CREATE OR REPLACE MATERIALIZED VIEW "items_view" AS SELECT *...
```


jeremyevans:
Thanks for the patch!  I think your implementation looks fine.  If we do eventually have another database that supports replacing materialized views, we can adjust at that point.

PR #2146, opened 2024-03-31
Add links to sequel-talk on the mail-archive.com

adam12:


jeremyevans:
Thanks for the patch!

PR #2150, opened 2024-04-05
Support SKIP LOCKED for MariaDB 10.6+.

simi:
Should be supported according to https://mariadb.com/kb/en/select/#skip-locked. I have tested locally and it works well. I wasn't able to find out any specs to cover this.

### DIFF
--- a/lib/sequel/adapters/shared/mysql.rb
+++ b/lib/sequel/adapters/shared/mysql.rb
@@ -926,9 +926,9 @@ module Sequel
         (type == :insert && db.mariadb? && db.adapter_scheme != :jdbc) ? (db.server_version >= 100500) : false
       end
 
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch!  There are integration tests for this, so I should be able to test tomorrow if it works. I think GitHub CI is MySQL and not MariaDB, so GitHub CI wouldn't be able to test this.

simi:
> Thanks for the patch! There are integration tests for this, so I should be able to test tomorrow if it works. I think GitHub CI is MySQL and not MariaDB, so GitHub CI wouldn't be able to test this.

I have tested locally with `version()` = `10.11.5-MariaDB-1:10.11.5+maria~ubu1804-log`. Indeed current GitHub Actions is MySQL only.

PR #2162, opened 2024-05-11
Clarify about `model_classes` in the example code

duongdominhchau:


jeremyevans:
Thanks for the patch!

PR #2173, opened 2024-06-01
Update testing guide with multithreaded transactional tests

janko:
People searching for testing advice will likely look at the testing guide, so I think it would be useful to have a section on browser tests there, to make the new `temporarily_release_connection` extension more discoverable.


jeremyevans:

PR #2191, opened 2024-07-23
Add string_agg support on sqlite >= 3.44

segiddins:
See https://www.sqlite.org/lang_aggfunc.html#group_concat for details

Distinct with a non-default separator is not supported due to a sqlite parser limitation


jeremyevans:
Thanks for the patch!  I'll try to ignore the unnecessary whitespace changes. :)  I'll get this tested and merged shortly.

PR #2195, opened 2024-07-25
 Allow proc default values that depend on other values

pedrocarmona:
Add ability to set default values that depend on other values from a given record.

Example: 
If slice belongs to cake with frostings, then allow to set frostings in cake slice by default.

```
class Slice < Sequel::Model
   many_to_one :cake
   plugin :defaults_setter
   
   def self.default_values
      {
        frosting: lambda { |slice| slice.cake&.frosting? },
      }
   end
 end
 ```
 
 ```
 Cake.create(frosting: true)
 Slice.new(cake_id: cake.id).frosting # => true
 ```

### DIFF
--- a/lib/sequel/plugins/defaults_setter.rb
+++ b/lib/sequel/plugins/defaults_setter.rb
@@ -43,13 +43,13 @@ module Sequel
     #   album.values # => {}
     #   album.a
     #   album.values # => {:a => Date.today}
-    # 
+    #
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch!  I'll test and merge later today.

pedrocarmona:
Thank you Jeremy 🙌

PR #2236, opened 2024-10-21
Use nested module format for Java package names

kalenp:
JRuby provides two formats for referring to java package [1].  Sequel has historically used a mix of both formats.  This consolidates on the "nested module" format (Java::JavaSql::Date) vs the java style format (java.sql.Date).  Apart from internal consistency, this format also presents as more idiomatic Ruby, since there are no invalid module names, such as Java::oracle.jdbc.driver.OracleDriver (`oracle` is not a valid module name).

[1] https://github.com/jruby/jruby/wiki/CallingJavaFromJRuby#referencing-java-classes-using-full-qualified-class-name

### DIFF
--- a/lib/sequel/adapters/jdbc.rb
+++ b/lib/sequel/adapters/jdbc.rb
@@ -20,7 +20,7 @@ module Sequel
 
     # Create custom NativeException alias for nicer access, and also so that
     # JRuby 9.2+ so it doesn't use the deprecated ::NativeException
-    NativeException = java.lang.Exception
+    NativeException = Java::JavaLang::Exception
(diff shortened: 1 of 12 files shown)


kalenp:
My use case is only affected by the OracleTypes declarations, but I figured it would be worth covering the rest for consistency sake.  I can reduce the scope of this if you like.

jeremyevans:
All tests passed, so I'll merge this shortly.

jeremyevans:
Cherry-picked at bba3e40d795ead1f34e3c48936db533d3ab331c8

PR #2242, opened 2024-11-04
Support using no_auto_parameterize to remove the insert limit

davekaro:
Before this change, if you enabled the pg_auto_parameterize extension, all inserts were limited to 40 rows at a time.

Now, you can use the `#no_auto_parameterize` method to turn off this behavior since inserts without parameterization can handle much higher rows at a time.


jeremyevans:
Thanks for the patch!

PR #2244, opened 2024-11-06
Update model docs for clarity

jonathanparrott:
I just learned of the `Model.first` alternative `Model[]`, but when reading the docs I initially attempted to use it as the docs show `Model.[]` which is a syntax error. My apologies if this is intentional, but it was confusing for me as a Sequel noob and thought I'd help clarify for others.


jeremyevans:
Thank you for the patch! Note that `Model.[]` is not a syntax error, it's a regular method call:

```ruby
  Model.first(id: 1)
  Model.[](1)
```

That being said, I can see where this could be confusing to users, so I will merge this.

PR #2249, opened 2024-11-10
MySQL and Postgres schema: retrieve columns' comments

Bahanix:
Following this discussion:
- https://github.com/jeremyevans/sequel/discussions/2248

Here is a PR to retrieve columns' comments (+ collation + privileges for MySQL).

As @jeremyevans shared, it may be more appropriate to build an extension, similar to @mpalmer's https://github.com/mpalmer/sequel-pg-comment. No hard feeling if the PR is rejected, I mostly pushed it because it fixes my use case, and it could help people who would come across the same issue.

Thank you for your help finding this solution!

### Usage

```ruby
DB.schema(table_name)
# Each column now has a "comment" key
```

#### MySQL vs Postgres adapters:

For other fields, MySQL adapter returns empty string when there is nothing; while Postgres adapter returns nil. I followed this pattern–but it means the behavior isn't the same for the two of them. I don't have an opinion for this, feel free to comment.

#### Adding vs retrieving comment

This PR is only about retrieving comment, hence hardcoded SQL queries in the tests.

### Tests

```sh
rake spec_mysql
rake spec_postgres
```

### DIFF
--- a/doc/reflection.rdoc
+++ b/doc/reflection.rdoc
@@ -59,6 +59,7 @@ The hash may also contain entries for:
 Database#schema takes a table symbol and returns column information in an array with each element being an array with two elements.  The first elements of the subarray is a column symbol, and the second element is a hash of information about that column.  The hash should include the following keys:
 
 :allow_null :: Whether NULL/nil is an allowed value for this column. Used by the Sequel::Model typecasting code.
+:comment:: The comment of the column (MySQL and PostreSQL).
 :db_type :: The type of column the database provided, as a string.  Used by the schema_dumper plugin for a more specific type translation.
 :default :: The default value of the column, as either a string or nil.  Uses a database specific format. Used by the schema_dumper plugin for converting to a ruby value.
 :primary_key :: Whether this column is one of the primary key columns for the table.  Used by the Sequel::Model code to determine primary key columns.
--- a/lib/sequel/adapters/shared/mysql.rb
+++ b/lib/sequel/adapters/shared/mysql.rb
@@ -578,6 +578,7 @@ module Sequel
           end
           row[:allow_null] = row.delete(:Null) == 'YES'
           row[:comment] = row.delete(:Comment)
+          row[:comment] = nil if row[:comment] == ""
           row[:default] = row.delete(:Default)
           row[:db_type] = row.delete(:Type)
           row[:type] = schema_column_type(row[:db_type])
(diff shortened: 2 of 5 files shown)


doc/reflection.rdoc
```diff
@@ -59,6 +59,7 @@ The hash may also contain entries for:
 Database#schema takes a table symbol and returns column information in an array with each element being an array with two elements.  The first elements of the subarray is a column symbol, and the second element is a hash of information about that column.  The hash should include the following keys:
 
 :allow_null :: Whether NULL/nil is an allowed value for this column. Used by the Sequel::Model typecasting code.
+:comment:: The comment of the column.
```
jeremyevans, line 62: This should be clear that it is MySQL/PostgreSQL only.

lib/sequel/adapters/shared/mysql.rb
```diff
@@ -577,8 +577,11 @@ def schema_parse_table(table_name, opts)
             row[:generated] = !!(extra.to_s =~ /VIRTUAL|STORED|PERSISTENT/i)
           end
           row[:allow_null] = row.delete(:Null) == 'YES'
+          row[:collation] = row.delete(:Collation)
```
jeremyevans, line 580: Collation should not be included, nor privileges.  This PR is only for comment support.

jeremyevans (changes requested):
I'm still on the fence about accepting this.  But one of the prerequisites would be that specs are added to test for expected behavior.

mpalmer:
You'll definitely get shoved down the slippery slope.  It seems somewhat inconsistent to provide in-core support for just reading *some* kinds of comments on *some* database engines, while needing an extension to get more comprehensive support.  A `mysql_comment` plugin would seem to be the more consistent approach, to provide access to whatever support MySQL has for such things.

Bahanix:
> Using nil instead of the empty string seems like the best choice.

Change pushed

To give you some context about the need, Sequel is a dependency of Langchainrb which is a dependency of a project I work on. The more info Langchainrb can retrieve from the DB, the better the underlying AI can perform – this is why reading is enough here.

jeremyevans:
> You'll definitely get shoved down the slippery slope. It seems somewhat inconsistent to provide in-core support for just reading _some_ kinds of comments on _some_ database engines, while needing an extension to get more comprehensive support. A `mysql_comment` plugin would seem to be the more consistent approach, to provide access to whatever support MySQL has for such things.

I agree that it seems somewhat inconsistent.  However, if this handles the needs of 80% of users who would like some comment support, considering the limited invasiveness, I think it should be worth it.  I would be open to expanding the same type of support by default on other databases, as long as it was of similar invasiveness, so the slippery slope there doesn't bother me too much.  In terms of reading other types of comments, and writing comments, that would be a bigger issue, but maybe it isn't a problem to consider that.

jeremyevans:
> > Using nil instead of the empty string seems like the best choice.
> 
> Change pushed

Looks like you need to update the MySQL adapter spec you added for the change.

Bahanix:
Woopsy, fixed

jeremyevans:
Squash merged at b59c0aaac2511b56f35d69ffc43bb99200a3f2d1

PR #2259, opened 2024-12-12
Slight clarification on how serialization plugin works

davetron5000:
Using the serialization plugin for the first time, I got the serializer/deserialzer backwards and the docs don't make this as clear as they could.

This has two changes:

* Document `format` for `serialize_attributes` to make it clear the ordering when using a custom serializer
* Change the example custom serializer from string reversal to Base64 encoding.  This makes the example slightly more verbose, but it's also now clear what is serializing vs not.

These are in separate commits in case one of these change is not desired while the other is

### DIFF
--- a/lib/sequel/plugins/serialization.rb
+++ b/lib/sequel/plugins/serialization.rb
@@ -37,7 +37,8 @@ module Sequel
     #
     #   # Register custom serializer/deserializer pair, if desired
     #   require 'sequel/plugins/serialization'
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch!

jeremyevans:
Squash merged at 0b9547d2cba0d30a0ebfcc8fe4bf7725d7842a5b

PR #2269, opened 2025-01-23
Fix confusing 'also' in server_block doc

artofhuman:
This PR fixing a bit of confusing statement, because there are only 2 queries.


jeremyevans:

PR #2279, opened 2025-02-26
Reduce queries made on startup when the `pg_enum` and `pg_array` plugins are enabled

nick96:
By not providing the scalar OID, the `pg_array` plugin will lookup each custom type individually. I've updated the `pg_enum` plugin to call `pg_array`'s `#register_array_type` with the scalar type's OID so it doesn't have to do the lookup. We've been running this in production today and we're no longer seeing our app make these a query per custom type.

Existing discussion: https://github.com/jeremyevans/sequel/discussions/2278


jeremyevans:
Everything looks good and works well, thank you for the patch!

PR #2288, opened 2025-03-15
Add `allow_by_default` option for `forbid_lazy_load` plugin

brandur:
This one was conceived and discussed more thoroughly over in #2286. The
basic idea is that the `forbid_lazy_load` plugin disables lazy loading
on specific model instances, but also broadly on any models loaded
through an association. The latter can be detrimental when trying to cut
over to eager loading incrementally and where it's hard to get an entire
codebase converted all at once.

Here, the `forbid_lazy_load` plugin picks up a new `allow_by_default`
option. When it's set to true, lazy loading can still be forbidden
normally on an instance-by-instance basis, but it's not forbidden
through associations. This should make incremental migrations to eager
loading with `forbid_lazy_load` somewhat easier to do.

### DIFF
--- a/lib/sequel/plugins/forbid_lazy_load.rb
+++ b/lib/sequel/plugins/forbid_lazy_load.rb
@@ -39,6 +39,9 @@ module Sequel
     #
     #   Album.first.artist # no error
     #
+    # This behavior of enabling +forbid_lazy_load+ automatically from dataset
+    # methods can be disabled using the plugin's +:allow_by_default+ option:
+    #
     # You can allow lazy loading associations for an instance that it
     # was previously forbidden for:
     # 
@@ -98,7 +101,17 @@ module Sequel
     #
     #   # Make the Album class support forbidding lazy load
     #   Album.plugin :forbid_lazy_load
+    #
+    #   # Let lazy loading be forbidden by object, but not automatically for any
+    #   # object loaded via dataset.
+    #   Album.plugin :forbid_lazy_load, allow_by_default: true
     module ForbidLazyLoad
+      def self.apply(model, opts=OPTS)
+        unless opts[:allow_by_default]
+          model.send(:dataset_extend, ForbidByDefault, :create_class_methods=>false)
+        end
+      end
+
       # Error raised when attempting to lazy load an association when
       # lazy loading has been forbidden.
       class Error < StandardError
@@ -179,7 +192,7 @@ module Sequel
         end
       end
 
-      module DatasetMethods
+      module ForbidByDefault
         # Mark model instances retrieved in this call as forbidding lazy loading.
         def each
           if row_proc
--- a/spec/extensions/forbid_lazy_load_spec.rb
+++ b/spec/extensions/forbid_lazy_load_spec.rb
@@ -1,10 +1,11 @@
 require_relative "spec_helper"
 
 describe "forbid_lazy_load plugin" do
-  before do
+  # opts are forwarded to `plugin :forbid_lazy_load`
+  def setup_model(opts={})
     @c = Class.new(Sequel::Model)
     @c.set_dataset Sequel::Model.db[:ts].with_fetch({:id=>2, :t_id=>3})
-    @c.plugin :forbid_lazy_load
+    @c.plugin :forbid_lazy_load, opts
     @c.columns :id, :t_id
     @c.many_to_one :t, :class=>@c, :key=>:t_id
     @c.one_to_many :ts, :class=>@c, :key=>:t_id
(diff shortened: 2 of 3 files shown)


lib/sequel/plugins/forbid_lazy_load.rb
```diff
@@ -179,7 +192,7 @@ def _load_associated_objects(opts, dynamic_opts=OPTS)
         end
       end
 
-      module DatasetMethods
+      module ForbidByDefault
```
brandur, line 195: Renamed as per https://github.com/jeremyevans/sequel/discussions/2286#discussioncomment-12512504, but should we maybe append "DatasetMethods" to the end of this to make it clear that these are specifically for that?

spec/extensions/forbid_lazy_load_spec.rb
```diff
@@ -15,312 +16,364 @@
...
   end
 
-  it "should not forbid lazy load if not set at instance level" do
-    @o1.t.must_equal @o2
-    @o1.ts.must_equal [@o2]
-    @o1.mtm_ts.must_equal [@o2]
-    @o1.otoo_t.must_equal @o2
-    @o1.oto_t.must_equal @o2
-  end
+  model_instance_specs = Module.new do
```
brandur, line 19: I extracted the core specs for model instances into this set and then included them in the two groups below.

spec/extensions/forbid_lazy_load_spec.rb
```diff
@@ -15,312 +16,364 @@
...
 
-  it "should forbid lazy load if retrieved by dataset via Dataset#each" do
-    o = @c.each{|x| break x}
-    proc{o.t}.must_raise Sequel::Plugins::ForbidLazyLoad::Error
-    proc{o.ts}.must_raise Sequel::Plugins::ForbidLazyLoad::Error
-    proc{o.mtm_ts}.must_raise Sequel::Plugins::ForbidLazyLoad::Error
-    proc{o.otoo_t}.must_raise Sequel::Plugins::ForbidLazyLoad::Error
-    proc{o.oto_t}.must_raise Sequel::Plugins::ForbidLazyLoad::Error
-  end
+  describe "default options" do
```
brandur, line 58: Most changes below are just the cases getting indented in by one so we could have two separate `describe` groups so `include model_instance_specs` works properly.

spec/extensions/forbid_lazy_load_spec.rb
```diff
@@ -15,312 +16,364 @@
...
+      5.times do
+        o = @c.where_each(:id=>1){|x| break x}
+        o.allow_lazy_load.must_be_same_as o
+        o.t.must_equal @o2
+        o.ts.must_equal [@o2]
+        o.mtm_ts.must_equal [@o2]
+        o.otoo_t.must_equal @o2
+        o.oto_t.must_equal @o2
+      end
+    end
```
brandur, line 377: Along with `include model_instance_specs`, I just inverted a couple of the core data set specs for this group. I didn't copy all of them because it feels too verbose/duplicative. This should be enough to prove the association behavior is working as expected.

spec/extensions/forbid_lazy_load_spec.rb
```diff
@@ -1,10 +1,11 @@
 require_relative "spec_helper"
 
 describe "forbid_lazy_load plugin" do
-  before do
+  # opts are forwarded to `plugin :forbid_lazy_load`
+  def setup_model(opts={})
     @c = Class.new(Sequel::Model)
     @c.set_dataset Sequel::Model.db[:ts].with_fetch({:id=>2, :t_id=>3})
-    @c.plugin :forbid_lazy_load
+    @c.plugin :forbid_lazy_load, **opts
```
jeremyevans, line 8: You should pass `opts` directly here (Sequel still supports Ruby 1.9).

spec/extensions/forbid_lazy_load_spec.rb
```diff
@@ -15,312 +16,364 @@
...
-    proc{@o1.otoo_t(:forbid_lazy_load=>true, :reload=>true)}.must_raise Sequel::Plugins::ForbidLazyLoad::Error
-    proc{@o1.oto_t(:forbid_lazy_load=>true, :reload=>true)}.must_raise Sequel::Plugins::ForbidLazyLoad::Error
-  end
+    include model_instance_specs
 
-  it "should not effect naked datasets" do
-    @c.naked.all.must_equal [{:id=>2, :t_id=>3}]
-    @c.naked.where_each(:id=>1){|x| break x}.must_equal(:id=>2, :t_id=>3)
-  end
+    it "should allow forbid lazy load if retrieved by dataset via Dataset#all" do
```
jeremyevans, line 347:
```suggestion
    it "should allow lazy load if retrieved by dataset via Dataset#all" do
```

jeremyevans (commented):
This looks good.  Just a couple minor fixes needed.

spec/extensions/forbid_lazy_load_spec.rb
brandur: Fixed!

spec/extensions/forbid_lazy_load_spec.rb
brandur: Got it! Fixed.

brandur:
Thx. Pushed fixes discussed and added a changelog entry.

jeremyevans:
Looks good.  Should be merging shortly.

lib/sequel/plugins/forbid_lazy_load.rb
```diff
@@ -39,6 +39,9 @@ module Plugins
     #
     #   Album.first.artist # no error
     #
+    # This behavior of enabling +forbid_lazy_load+ automatically from dataset
+    # methods can be disabled using the plugin's +:allow_by_default+ option:
```
brandur, line 43: @jeremyevans Ah doh, I just noticed that the colon at the end of this line should be a period. I think I'd intended to put an example under it originally, then decided to put it in the examples section below instead. Probably not worth a PR over, but you may want to do a ninja commit to master or something.

jeremyevans: Thanks for the heads up. I pushed a commit to fix this.

PR #2297, opened 2025-03-30
Add UTF-8 support for ODBC adapter by use_odbc_utf8 option

mnmandahalf:
# Background

There have been some problems with the ODBC adapter handling UTF-8 strings, including problems allocating buffers to store the strings. Details are described in the following blog.

https://mnmandahalf.hatenablog.com/entry/2025/03/31/003137

# What this PR is doing

This PR implements the `:use_odbc_utf8` option for the ODBC adapter that allows using the `ODBC_UTF8` module instead of the standard `ODBC` module.  (see: http://www.ch-werner.de/rubyodbc/README)

This enables proper handling of UTF-8 encoded data when communicating with the ODBC driver.

Changes include:
- Dynamic module selection based on `:use_odbc_utf8` option
- Support for both `ODBC` and `ODBC_UTF8` error types
- Proper type conversion for date/time/bit values from both modules
- Updated documentation explaining the new option

# Other considerations
I considered switching by `encoding` option, but avoided it because it would be an implicit change and would have a significant impact on error handling based on the content of error messages.

The behavior was checked in a local environment with the Snowflake ODBC driver, but the appropriate spec file could not be found.

### DIFF
--- a/doc/opening_databases.rdoc
+++ b/doc/opening_databases.rdoc
@@ -300,6 +300,8 @@ The :host and :port options are not respected. The following additional options 
 :db_type :: Can be specified as 'mssql', 'progress', or 'db2' to use SQL syntax specific to those databases.
 :drvconnect :: Can be given an ODBC connection string, and will use ODBC::Database#drvconnect to
                do the connection.  Typical usage would be: <tt>Sequel.odbc(drvconnect: 'driver={...};...')</tt>
+:use_odbc_utf8 :: When set to true, uses the ODBC_UTF8 module instead of the standard ODBC module.
+                  This enables proper handling of UTF-8 encoded data when communicating with the ODBC driver.
 
 === oracle 
 
--- a/lib/sequel/adapters/odbc.rb
+++ b/lib/sequel/adapters/odbc.rb
@@ -1,6 +1,7 @@
 # frozen-string-literal: true
 
 require 'odbc'
+require 'odbc_utf8'
 
 module Sequel
   module ODBC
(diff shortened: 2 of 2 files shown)


jeremyevans:
Thank you for submitting a patch!

I ran this through the integration tests with the odbc/mssql adapter, and after a minor change to that, there was only 1 test failure.  `Dataset#update` appears to return `-1` instead of `0` if no rows are modified in UTF-8 mode.  This seems like a bug in the ruby-odbc code in UTF8 mode, in the `stmt_nrows` function in the `odbc.c` file.

I'm not sure we actually need this, though.  With the current odbc adapter, if you make no changes, and just require `odbc_utf8` before Sequel, you get the same behavior as with this patch.  This is due to the design of ruby-odbc gem, which defines different constants for requires of `odbc` and `odbc_utf8`, based on which is required first.  The only reason to have this patch is if we wanted to support both UTF8 and non-UTF8 ODBC connections in the same process.  Do you actually need to do that?  Unless there is a strong need for it, I don't think the additional complexity is worth it.

jeremyevans:
I'm going to close this now.  If you actually need to handle both UTF-8 and non-UTF-8 ODBC connections in the same process, please respond.

mnmandahalf:
@jeremyevans
Thanks for the detailed feedback and running those tests! 
I agree that we should avoid adding complexity unless there's a compelling reason to accommodate both types of connections. 
And it's actually very restricted situation.
Thanks again for your analysis.

PR #2300, opened 2025-04-12
Support paged_updates with multiple FROM tables

davekaro:
When you need to update a column using the value in another table, you need the primary key to be qualified. Discussed in https://github.com/jeremyevans/sequel/discussions/2299

### DIFF
--- a/lib/sequel/plugins/paged_operations.rb
+++ b/lib/sequel/plugins/paged_operations.rb
@@ -159,7 +159,7 @@ module Sequel
 
           case pk = model.primary_key
           when Symbol
-            Sequel.identifier(pk)
+            Sequel[model.table_name][pk]
           when Array
             raise Error, "cannot use #{meth} on a model with a composite primary key"
           else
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch!

jeremyevans:
I think I should do additional work in this area before the next release.   While this PR works, it seems odd the sometimes the table is qualified and sometimes it isn't, and this issue seems likely to affect `paged_each` in addition to the paged_operations plugin.  I'll see if I can make it consistent.  If you are joining to another table, it makes sense to qualify it.  If you aren't joining to another table, seems unnecessary to qualify it.

jeremyevans:
I changed things so that paged_each/paged_operations use a qualified primary key for a joined dataset.  I chose not to use this PR because I would have had to back out all of the spec changes, and I didn't want the extra churn.  I listed you as a co-author on the commit, since the idea was yours.

davekaro:
Sounds good, all makes sense. Thanks!

PR #2305, opened 2025-05-14
Add MariaDB to CI

grooverdan:
As  a quick test to ensure that MariaDB is remaining true and compatible to our join user base, lets test it.

I considered the jdbc-mariadb driver however that seems in need of an update.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,13 @@ jobs:
           MYSQL_DATABASE: sequel_test
         ports: ["3306:3306"]
         options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+      mariadb:
+        image: mariadb:latest
+        env:
+          MARIADB_ROOT_PASSWORD: root
+          MARIADB_DATABASE: sequel_test
+        ports: ["3307:3306"]
+        options: --health-cmd="healthcheck.sh --connect --innodb_initialized" --health-interval=10s --health-timeout=5s --health-retries=3
     strategy:
       fail-fast: false
       matrix:
--- a/Rakefile
+++ b/Rakefile
@@ -157,16 +157,23 @@ end
 task :spec_ci=>[:spec_core, :spec_model, :spec_plugin, :spec_core_ext] do
   mysql_host = "localhost"
   pg_database = "sequel_test" unless ENV["DEFAULT_DATABASE"]
+  mysql_jdbc = "&allowPublicKeyRetrieval=true"
 
   if ENV["MYSQL_ROOT_PASSWORD"]
     mysql_password = "&password=root"
     mysql_host= "127.0.0.1:3306"
   end
 
+  if ENV["MARIADB_ROOT_PASSWORD"]
+    mysql_password = "&password=root"
+    mysql_host = "127.0.0.1:3307"
+    mysql_jdbc = ""
+  end
+
   if defined?(RUBY_ENGINE) && RUBY_ENGINE == 'jruby'
     ENV['SEQUEL_SQLITE_URL'] = "jdbc:sqlite::memory:"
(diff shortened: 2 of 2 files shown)


jeremyevans:
Thank you for submitting a patch. There is a benefit and a cost to adding databases to the CI, and considering the trade-offs, I don't think it is worth it to test both MySQL and MariaDB in CI. Sequel supports numerous databases that aren't tested in GitHub Actions CI. If we wanted to add an additional database to CI, it would likely be Microsoft SQL Server.  MariaDB is the primary MySQL-like database I test against locally, so having MySQL in CI is good for catching MySQL-specific issues that do not apply to MariaDB.

grooverdan:
Hi @jeremyevans, thanks for your prompt response.

I'm grateful for your local testing of MariaDB. Which MariaDB version do you test? There's an aspect of earliest/latest that could provide benefits beyond what you don't test locally I assume.

While MariaDB's compatibility with a MySQL that existed when if forked is quite likely there has been plenty of scope for divergence in the 15 since it forked.

On CI reduction options:

1.

The easiest is if MariaDB is included in only one or two Ruby versions of the ci matrix.

Seems even with this PR I missed the step in the workflow that would set and trigger the environment variable trigger.

2.

A cron as mentioned is certainly an option some other projects use. By your release tags it seems you release more than monthly so having some validation before then is probably useful? weekly? The full range of ruby runtimes, not my strongest language base, but I suspect only a small number ruby runtimes contributes to any addition testing coverage in a MariaDB test run.

An earliest/ latest MariaDB supported version could double check the earliest is maintained and the latest isn't introducing regressions. 

What would be a minimal acceptable set?

3.

Triggered. An option used by the MariaDB Operator is that when MariaDB's CI produces a new test container from the finished developer works per release series, it calls a github action to run its test suite against the container.

Its config workflow is: https://github.com/mariadb-operator/mariadb-operator/blob/main/.github/workflows/test-image.yml

It involves a github token swap - one to me to allow me to trigger the CI, and one to you so the github action can tell MariaDB on zulip that something incompatible happened. This way its not on you to investigate incompatibilities. This gets called about 1-3 times per day when developers commit changes to the release branches.

So not exactly minimal, but timely and out of band to PR and commit test runs and making sure the next MariaDB version is compatible rather than one or both of our project scrambling after release.

4.

Run per request was mentioned, I haven't looked up how exactly that is done. Maybe its like above (3) where a `gh` client does the trigger. Or a named github branch?


So yes, I'm happy to work on any of these. What's your preference?

jeremyevans:
Sequel development is mostly done on OpenBSD-current, and the MariaDB version would follow whatever MariaDB version OpenBSD supports (currently 11.4.5).  At one point, I was testing multiple MariaDB versions, back to MariaDB 10.0, but I ran into build issues after an OpenBSD upgrade a while back, so I no longer do so.

I certainly understand that there has been substantial divergence between MariaDB and MySQL.  However, Sequel still uses the mysql/mysql2/trilogy adapters for both MariaDB and MySQL.  Places where behavior is different between the two are generally handled with conditionals inside the shared mysql adapter. 

I suppose swapping MySQL for MariaDB in the CI matrix for some Ruby versions could be considered, as long as it doesn't complicate the workflow too much.

Sequel releases are monthly, on the first of the month. In terms of a `cron` trigger, maybe run around the 25th of the month?  That still gives some time to fix issues if any show up. In terms of Ruby versions, the highest value is testing the lowest supported (Ruby 1.9) and highest supported (currently Ruby 3.4). 

I don't test prerelease versions of other databases in CI, so I don't I want to do so for MariaDB.  

I think the run per request uses a `workflow_dispatch` trigger.  It can be run manually through the GitHub UI.  You can have multiple triggers, such as both `cron` and `workflow_dispatch`.  That would allow for earlier CI testing of changes considered more likely to cause regressions.

PR #2319, opened 2025-07-23
Add class_table_inheritance_constraint_validations plugin

sterlzbd:
This plugin updates constraint_validations to parse validations from all ancestor tables in a class_table_inheritance hierarchy. This allows subclasses to automatically inherit constraint_validations defined in their superclasses.

more discussion can be found here: https://github.com/jeremyevans/sequel/discussions/2316

I tried to match the current style of docs and tests but I'm not that familiar with rspec (mostly use minitest).


sterlzbd:
I think I see the issue. I believe my new tests are competing with the main `constraint_validations` plugin tests over the `constraint_validations_table`.

jeremyevans:
Thank you for your continued work on this. I'm running this through some final testing, and it should be merged later today.

sterlzbd:
Thanks and thank you for all your work on Sequel!

PR #2360, opened 2026-04-22
test: fix `must_raise` assertions on minitest 6.0.5

chadlwilson:
On Minitest 6.0.5 these assertions now throw errors if passed a regex into `must_raise`

Don't think Minitest ever strictly supported this syntax as originally intended (noted at https://github.com/minitest/minitest/issues/1068#issuecomment-4263808393 ) since the string is a custom error for the assertion, not an expectation on the message.

e.g
```
  1) Error:
Database.after_initialize#test_0003_should raise an error if registration is called without a block:
TypeError: class or module required for rescue clause. Got [Sequel::Error, /must provide block/i]
    spec/core/database_spec.rb:3292:in 'block in <main>'
    org/jruby/RubyArray.java:2093:in 'each'
    org/jruby/RubyArray.java:2892:in 'map'
```

https://github.com/minitest/minitest/blob/master/History.rdoc#605--2026-04-20

This change fixes the assertions for all `must_raise` I could find. CI passing on my fork.

### DIFF
--- a/spec/adapters/postgres_spec.rb
+++ b/spec/adapters/postgres_spec.rb
@@ -844,9 +844,10 @@ describe "PostgreSQL temporary table/view support" do
   end
 
   it "should not allow to pass both :temp and :unlogged" do
-    proc do
+    err = proc do
       @db.create_table(:temp_unlogged_dolls, :temp => true, :unlogged => true){text :name}
-    end.must_raise(Sequel::Error, "can't provide both :temp and :unlogged to create_table")
+    end.must_raise(Sequel::Error)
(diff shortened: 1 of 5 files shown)


jeremyevans:
Thanks for the patch! This isn't the first time Minitest has broken compatibility in a tiny release, and I doubt it will be the last.

chadlwilson:
> Thanks for the patch! This isn't the first time Minitest has broken compatibility in a tiny release, and I doubt it will be the last.

To be fair to them, it has highlighted that the `must_raise` assertions on the messages were not correctly applied since the migration to minitest, so maybe we should thank them? 🙏

PR #2361, opened 2026-04-22
build: add JRuby 10.1 to test matrix

chadlwilson:
https://www.jruby.org/2026/04/21/jruby-10-1-0-0

(depends on #2360)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        ruby: [ "2.0.0", 2.1, 2.3, 2.4, 2.5, 2.6, 2.7, "3.0", 3.1, 3.2, 3.3, 3.4, "4.0", jruby-9.3, jruby-9.4, jruby-10.0 ]
+        ruby: [ "2.0.0", 2.1, 2.3, 2.4, 2.5, 2.6, 2.7, "3.0", 3.1, 3.2, 3.3, 3.4, "4.0", jruby-9.3, jruby-9.4, jruby-10.0, jruby-10.1 ]
         include:
           - { os: ubuntu-22.04, ruby: "1.9.3" }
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch! I rebased merged #2360, and I think GitHub has problems figuring out the first commit has already been applied. Can you please rebase this on master? I'm good to merge this as long as CI passes. I thought I could do the rebase automatically on GitHub, but maybe it doesn't show up because this is marked as a draft (not sure, just a guess).

chadlwilson:
Done. I kept it as draft as I was intending to rebase after you'd merged :-)

chadlwilson:
You keep a relatively impressive historic build matrix there. I am mildly terrified for the security posture of anyone still using JRuby 9.2 or maybe even 9.3 - even ignoring all the CRubies 😅

jeremyevans:
> You keep a relatively impressive historic build matrix there. I am mildly terrified for the security posture of anyone still using JRuby 9.2 or maybe even 9.3 - even ignoring all the CRubies 😅

Thank you. Mostly backwards compatibility with old Rubies is kept because it hasn't been that difficult to maintain support, and it allows someone with a legacy program on a very old (J)Ruby version to test with the current Sequel version, and then jump all the way to the current (J)Ruby version using the same Sequel version, compared to a lot of other libraries, where you probably need multiple smaller jumps. Sequel actually supports older JRuby versions, I think back to JRuby 1.6 using --1.9, but CI doesn't pass for those, so that's why they were dropped from CI.

chadlwilson:
These days between JRuby / Rails / Ruby language version/ Java version / misc non-default gem versions I've ended up in quite the quagmire trying to move forward with upgrades. And that's for a relarively simple application.

Even though I don't use sequel there (mainly looking at this to help out JRuby core) it's nice when not everything among ones dependencies is ratcheting compatibility over time and mercilessly dropping compat the day after EOL of a dependency.

PR #2366, opened 2026-05-08
Adds expiration for existing connections when added to a db

cooljacob204:
When connection expiration is added to a db it does not register existing connections. This loops through all available and allocated connections and sets an expiration value for them on initial load.

One alternative approach I though of was to instead lazy apply the expiration on connection checkout if it's missing. Might result in cleaner code but it would result in an extra check every time a connection is checked out.

Related discussion: https://github.com/jeremyevans/sequel/discussions/2365


jeremyevans:
Thanks for working on this! I'll review and test this shortly.

PR #2374, opened 2026-07-08
Fix specs on TruffleRuby and restore it on CI

andrykonchin:
Wondering if it is OK to add TruffleRuby back on CI.

I have fixed a couple of issues that lead to specs failures and were caused by minor differences between CRuby and TruffleRuby.

cc @eregon

### DIFF
--- a/lib/sequel/database/schema_generator.rb
+++ b/lib/sequel/database/schema_generator.rb
@@ -17,7 +17,9 @@ module Sequel
             if RUBY_VERSION >= "3.2"
             # :nocov:
               caller_loc = Thread.each_caller_location do |loc|
-                break loc unless loc.path == __FILE__
+                # Skip core library methods implemented in Ruby
+                path = loc.path
+                break loc unless path == __FILE__ || (path && path.start_with?('<internal:'))
               end
               caller_loc &&= "#{caller_loc.path}:#{caller_loc.lineno}: "
             end
--- a/spec/extensions/caller_logging_spec.rb
+++ b/spec/extensions/caller_logging_spec.rb
@@ -39,7 +39,10 @@ describe "caller_logging extension" do
     @log_stream.rewind
     lines = @log_stream.read.split("\n")
     lines.length.must_equal 1
-    lines[0].must_match(/ \(source: #{__FILE__}:#{line}:in [`']block.+\) SELECT \* FROM items\z/)
+
+    # TruffleRuby names block frames after the compiled/optimized test method name
+    # (e.g. test_0003_...) instead of 'block'.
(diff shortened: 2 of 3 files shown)


lib/sequel/database/schema_generator.rb
```diff
@@ -17,7 +17,16 @@ def _merge_column_options(defaults, opts)
...
             # :nocov:
               caller_loc = Thread.each_caller_location do |loc|
-                break loc unless loc.path == __FILE__
+                if RUBY_ENGINE == 'truffleruby'
+                  # TruffleRuby implements Hash#merge! in Ruby rather than C, which
+                  # introduces internal standard library frames (starting with <internal:)
+                  # into the backtrace when the block is yielded. We skip these frames
+                  # to find the correct user caller location.
+                  path = loc.path
+                  break loc unless path == __FILE__ || (path && path.start_with?('<internal:'))
```
jeremyevans: I think it would be good to have this as the default behavior, so we don't need the `RUBY_ENGINE` check. WDYT?

spec/extensions/caller_logging_spec.rb
```diff
@@ -39,7 +39,13 @@ def exec_sql(sql)
     @log_stream.rewind
     lines = @log_stream.read.split("\n")
     lines.length.must_equal 1
-    lines[0].must_match(/ \(source: #{__FILE__}:#{line}:in [`']block.+\) SELECT \* FROM items\z/)
+    if RUBY_ENGINE == 'truffleruby'
+      # TruffleRuby names block frames after the compiled/optimized test method name
+      # (e.g. test_0003_...) instead of 'block'.
+      lines[0].must_match(/ \(source: #{__FILE__}:#{line}:in [`']test_.+\) SELECT \* FROM items\z/)
+    else
+      lines[0].must_match(/ \(source: #{__FILE__}:#{line}:in [`']block.+\) SELECT \* FROM items\z/)
```
jeremyevans: Likewise, instead of `RUBY_ENGINE` check, use `(test_|block)` so it matches either.

jeremyevans (commented):
Thank you for the patch! No objects in principle to testing TruffleRuby in CI if it passes.

lib/sequel/database/schema_generator.rb
eregon: Yes definitely, this is designed to be the same pattern as in CRuby (https://gist.github.com/eregon/912e6359e83781c5fa1c638d3768c526)

I'd update the comment to:
```
# Skip core library methods implemented in Ruby
```

spec/extensions/caller_logging_spec.rb
andrykonchin: Yeah, makes sense. Done.

PR #2383, opened 2026-08-17
Recognize disconnects in the mysql and mysql2 adapters using client error codes

sdalu:
The disconnect error messages both adapters check are the wordings used by libmysqlclient.  MariaDB Connector/C omits the "MySQL" prefix from three of them, and ruby-mysql raises "Lost connection to server during query" when a connection is lost mid-query, so with either driver a lost connection was raised as a plain Sequel::DatabaseError and the dead connection was returned to the connection pool instead of being removed.

The numeric client error codes do not differ between client libraries, so check those in addition to the SQL state and the error message.  Message matching is still needed for errors raised by the driver itself, which have no client error code.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 === master
 
-* Recognize disconnects using client error codes in the mysql and mysql2 adapters, fixing disconnect detection when the client library or driver does not use libmysqlclient's error messages (sdalu) (#2382)
+* Recognize additional disconnect errors in the mysql and mysql2 adapters (sdalu) (#2382)
 
 * Deprecate Sequel.require (jeremyevans)
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 === master
 
+* Recognize disconnects using client error codes in the mysql and mysql2 adapters, fixing disconnect detection when the client library or driver does not use libmysqlclient's error messages (sdalu) (#2382)
+
 * Deprecate Sequel.require (jeremyevans)
 
 * Add ndims, position, positions, prepend, sample, shuffle, and trim to pg_array_ops extension (jeremyevans)
--- a/lib/sequel/adapters/utils/mysql_mysql2.rb
+++ b/lib/sequel/adapters/utils/mysql_mysql2.rb
@@ -31,6 +31,7 @@ module Sequel
           2014, # CR_COMMANDS_OUT_OF_SYNC
           2055, # CR_SERVER_LOST_EXTENDED
         ].freeze
+        private_constant :MYSQL_DISCONNECT_ERROR_CODES
 
         # Support stored procedures on MySQL
         def call_sproc(name, opts=OPTS, &block)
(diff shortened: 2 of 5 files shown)


jeremyevans:
Thank you for the patch. I'll try to review and merge later today.

lib/sequel/adapters/utils/mysql_mysql2.rb
```diff
@@ -22,7 +22,20 @@ module DatabaseMethods
         END
         # Error messages for mysql and mysql2 that indicate the current connection should be disconnected
         MYSQL_DATABASE_DISCONNECT_ERRORS = /\A#{Regexp.union(disconnect_errors)}/
-       
+
+        # Client error codes for mysql and mysql2 that indicate the current connection
+        # should be disconnected.  Unlike the related error messages, these do not
+        # differ between client libraries.  2027 (CR_MALFORMED_PACKET) is deliberately
+        # not included, as it indicates a protocol error and not necessarily a lost
+        # connection.
```
jeremyevans: From the constant name, the purpose is obvious. Might be worth including the 2027 information, since that isn't obvious.

```suggestion
        # 2027 (CR_MALFORMED_PACKET) deliberately not included (protocol error, not connection error)
```

spec/adapters/mysql_spec.rb
```diff
@@ -1059,6 +1059,42 @@ def convert_tinyint_to_bool?() false end #mysql2
...
+    it "should handle errors raised by the driver without a client error code as disconnects" do
+      disconnect_error?("MySQL client is not connected", nil).must_equal true
+    end
+
+    it "should not handle server errors as disconnects" do
+      disconnect_error?("Table 'test.foo' doesn't exist", 1146).must_equal false
+    end
+  end
+end
+
```
jeremyevans: These specs should be removed. The adapter tests are designed to run without mocking.

```suggestion
```

CHANGELOG
```diff
@@ -1,5 +1,7 @@
 === master
 
+* Recognize disconnects using client error codes in the mysql and mysql2 adapters, fixing disconnect detection when the client library or driver does not use libmysqlclient's error messages (sdalu) (#2382)
```
jeremyevans:
```suggestion
* Recognize additional disconnect errors in the mysql and mysql2 adapters (sdalu) (#2382)
```

lib/sequel/adapters/utils/mysql_mysql2.rb
```diff
@@ -22,7 +22,20 @@ module DatabaseMethods
...
+        MYSQL_DISCONNECT_ERROR_CODES = [
+          2002, # CR_CONNECTION_ERROR
+          2006, # CR_SERVER_GONE_ERROR
+          2013, # CR_SERVER_LOST
+          2014, # CR_COMMANDS_OUT_OF_SYNC
+          2055, # CR_SERVER_LOST_EXTENDED
+        ].freeze
+
```
jeremyevans, line 35: Let's make this constant private:

```suggestion
        private_constant :MYSQL_DISCONNECT_ERROR_CODES

```

jeremyevans (commented):
Looks good, only a few minor changes requested. If you'd like me to handle the changes, please let me know.

sdalu:
I applied the suggestion, do I need to do something else? I'm happy with it if you can handle the changes.

jeremyevans:
Nothing else needed, this looks good. I'll do some final testing and merge later today.

jeremyevans:
Squash merged at de98abae1e9170b00b51d6daca0e21bc6bb95b23

PR #2385, opened 2026-08-25
Add :port to Sequel.connect general options doc

jlmuir:
I noticed that `:port` is not listed in the

[General connection options](https://sequel.jeremyevans.net/rdoc/files/doc/opening_databases_rdoc.html#label-General+connection+options)

section of the "Connecting to a database" documentation.  This PR adds it.  Because `:host` is there, I'm fairly sure `:port` should be there too, but if I'm wrong about that, then of course just reject this PR.  I see references to the `:port` option in other parts of the same document, and a quick search of the source code finds references to it, so I think it's considered to be a general connection option, but that's about the extent of my looking into this.


jeremyevans (commented):
Thank you for the patch.  There are some adapters that don't support `:port`, but there are also adapters that don't support `:host`, and we don't call those out in this documentation, so no need to do so for port.

PR #2388, opened 2026-08-29
Reformat the rdoc Hanna documentation to match the existing website style

danielhoey:
The links under the "General Info, Guides, Examples, and Tutorials" section of /documentation.html on the website are changed to point to reformatted version of the corresponding rdoc files. The new formatting means these pages match the website style. I've referred to these pages as 'guides' in the Rakefile to distinguish them from the other documentation, maybe they could just be called 'docs'.

The reformatting relies upon the rdoc pages being produced using the Hanna template. The body of these pages is copied and new pages rendered within the existing layout.html.erb and placed in /doc. Links in the website headers are changes from relative to absolute to so that they work in the /doc subpath. Relative links in the new page are updated to either point to the absolute path to the rdoc file or to the corresponding new page if it exists.

The existing rake tasks that make the rdocs and website have been changed to file tasks so they only run if a source file has changed, that way I can make the 'website_guides' task depending the main rdoc task without need to rebuild rdoc from scratch everytime when I was developing.

The first heading in the README was changed to a top level heading so that the reformatting worked the same way on every page.


jeremyevans:
Thank you for working on this. I'll start reviewing this right away.

www/layout.html.erb
```diff
@@ -2,36 +2,36 @@
...
   
   <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
   <meta name="keywords" content="ruby, sequel, sql, toolkit, orm, plugins, documentation" />
   <meta name="description" content="Ruby Sequel is a lightweight database toolkit for Ruby." />
 
-  <link rel="icon" type="image/icon" sizes="32x32" href="images/favicon.ico">
-  <link href="css/application.css" media="screen" rel="stylesheet" type="text/css" />
-  <link href="css/<%= title %>.css" media="screen" rel="stylesheet" type="text/css" />
+  <link rel="icon" type="image/icon" sizes="32x32" href="/images/favicon.ico">
+  <link href="/css/application.css" media="screen" rel="stylesheet" type="text/css" />
```
jeremyevans: The problem with using absolute paths here is that it it requires a web server to work correctly. I'm guessing the links will be broken if you open the file locally. This should be fixable by using an approach like `<%= link_prefix %>css/application.css`, and set `link_prefix` to `'../'` when rendering the pages in the `doc` directory.

www/layout.html.erb
```diff
@@ -2,36 +2,36 @@
 <html lang="en">
```
jeremyevans: The problem with using absolute paths in the layout is that it requires a web server to work correctly. I'm guessing the links will be broken if you open the file locally in a browser.

Here's an approach that may work:

```suggestion

  <% link_prefix = "../" if defined?(stylesheet) && stylesheet == "guide" %>
```

jeremyevans:
I squash merged this at 27f1b112f04879bd3e9c810580a068f40e369deb. I made changes to use relative paths instead of absolute paths, so that you can still view and use the site directly after generating it, without running a web server.

